PR #1, opened 2022-04-08
CI: Import and adapt scripts from osmt

aehyvari:


MasoudAsadzade:
Nice, LGTM!

.circleci/config.yml
```diff
@@ -0,0 +1,148 @@
...
+                FLAGS: -Wall -Wextra -Werror
+
+        steps:
+            - checkout
+            - run:
+                name: apt-get update
+                command: sudo apt-get update
+            - run:
+                name: install dependencies
+                command: sudo apt-get install libgmp-dev libedit-dev bison flex cmake libubsan0 python3-pip cython
```
blishko, line 20: `libedit-dev` should not be needed for Golem

.circleci/config.yml
```diff
@@ -0,0 +1,148 @@
...
+
+        steps:
+            - run:
+                name: Install the environment
+                command: |
+                    yum -y install centos-release-scl
+                    yum -y install devtoolset-8
+                    yum -y install gmp-devel
+                    yum -y install gmp-static
+                    yum -y install libedit-devel
```
blishko, line 69: Same here

.circleci/config.yml
```diff
@@ -0,0 +1,148 @@
...
+                    wget --no-verbose "https://github.com/westes/flex/files/981163/flex-2.6.4.tar.gz"
+                    tar -xvzf flex-2.6.4.tar.gz
+                    cd flex-2.6.4
+                    echo "./configure --prefix=/usr/local; make -j4; make install" | scl enable devtoolset-8 bash
+                    cd ..
+                    rm -rf flex-*
+                    flex --version
+            - checkout
+            - run:
+                name: Set up model validation environment
```
blishko, line 99: This step should be removed.

.circleci/config.yml
```diff
@@ -0,0 +1,148 @@
...
+            - run:
+                name: Debug build gcc under devtoolset-8
+                command: |
+                    cat ./ci/run_commands.sh |scl enable devtoolset-8 bash
+                environment:
+                    CMAKE_BUILD_TYPE: Debug
+            - run:
+                name: Release build gcc under devtoolset-8
+                command: |
+                    cat ./ci/build_maximally_static.sh |scl enable devtoolset-8 bash
```
blishko, line 110: This script is missing now. Maybe it would make sense to have it as well?

blishko (changes requested):
@aehyvari, thanks for preparing this. I think it will need a few more adjustments. Do you want to do that or should I take over?

aehyvari:
If you feel like fixing them, I don't want to slow you down. All sounds
good to me!

.circleci/config.yml
blishko: I have added the script, but this feature was added to OpenSMT after the last release, so we can enable it only after the next release of OpenSMT.

blishko (approved):

PR #3, opened 2022-04-10
CI: remove unneeded rpms

aehyvari:
These two seem not to be needed.  I'm guessing that `python3-pip` has some version of `python3` as a dependency already.


blishko (approved):
LGTM!

PR #13, opened 2023-02-10
TPA transition system tree implementation

BritikovKI:
Implementation of the possible branching of the TSs.

### DIFF
--- a/src/TransformationUtils.cc
+++ b/src/TransformationUtils.cc
@@ -83,10 +83,6 @@ std::unique_ptr<TransitionSystem> toTransitionSystem(ChcDirectedGraph const & gr
     return ts;
 }
 
-//void strongconnect (SymRef v){
-//
-//}
-
 bool strongConnection(
         std::set<int>& visitedVertices,
         std::set<int>& verticesOnStack,
@@ -99,25 +95,23 @@ bool strongConnection(
     auto const & outEdges = graphRepresentation.getOutgoingEdgesFor(node);
 
     for (EId eid : outEdges) {
-        if(size(outEdges) <= 1){
+        if (size(outEdges) <= 1) {
             return false;
         }
-        if(graph.getTarget(eid) != node){
+        if (graph.getTarget(eid) != node) {
             auto nextVertice = graph.getTarget(eid);
             if(visitedVertices.find(nextVertice.x) == visitedVertices.end()){
-                bool loop_found = strongConnection( visitedVertices,
-                                                    verticesOnStack,
-                                                    graphRepresentation,
-                                                    graph,
-                                                    nextVertice);
-                if(loop_found){
+                bool loop_found = strongConnection(visitedVertices,
+                                                   verticesOnStack,
+                                                   graphRepresentation,
+                                                   graph,
+                                                   nextVertice);
+                if (loop_found) {
                     return true;
                 }
-            } else if (verticesOnStack.find(nextVertice.x) != verticesOnStack.end()){
+            } else if (verticesOnStack.find(nextVertice.x) != verticesOnStack.end()) {
                 return true;
             }
-        } else {
-            continue;
         }
     }
 
@@ -134,20 +128,20 @@ bool TarjanLoopDetection(ChcDirectedGraph const & graph) {
     std::set<int> verticesOnStack;
 
 
-    for(uint i = 1; i < vertices.size()-1; i++){
-        if(visitedVertices.find(vertices[i].x) == visitedVertices.end()){
+    for (uint i = 1; i < vertices.size()-1; i++) {
+        if (visitedVertices.find(vertices[i].x) == visitedVertices.end()) {
             bool loop_detected = strongConnection(visitedVertices,
                                                   verticesOnStack,
                                                   graphRepresentation,
                                                   graph,
                                                   vertices[i]);
-            if (loop_detected) return true;
+            if (loop_detected) { return true; }
         }
     }
     return false;
 }
 
-bool isTransitionSystemChain(ChcDirectedGraph const & graph) {
+bool isTransitionSystemDAG(ChcDirectedGraph const & graph) {
     if (graph.getVertices().size() < 3) { return false; }
     auto graphRepresentation = AdjacencyListsGraphRepresentation::from(graph);
     auto vertices = reversePostOrder(graph, graphRepresentation);
--- a/src/TransformationUtils.cc
+++ b/src/TransformationUtils.cc
@@ -4,6 +4,7 @@
 
 #include "TransformationUtils.h"
 #include "QuantifierElimination.h"
+#include <set>
 
 bool isTransitionSystem(ChcDirectedGraph const & graph) {
     auto graphRepresentation = AdjacencyListsGraphRepresentation::from(graph);
@@ -82,25 +83,86 @@ std::unique_ptr<TransitionSystem> toTransitionSystem(ChcDirectedGraph const & gr
     return ts;
 }
 
+//void strongconnect (SymRef v){
+//
+//}
+
+bool strongConnection(
+        std::set<int>& visitedVertices,
+        std::set<int>& verticesOnStack,
+        AdjacencyListsGraphRepresentation& graphRepresentation,
+        ChcDirectedGraph const & graph,
+        SymRef node)
+        {
+    visitedVertices.insert(node.x);
+    verticesOnStack.insert(node.x);
+    auto const & outEdges = graphRepresentation.getOutgoingEdgesFor(node);
+
+    for (EId eid : outEdges) {
+        if(size(outEdges) <= 1){
+            return false;
+        }
+        if(graph.getTarget(eid) != node){
+            auto nextVertice = graph.getTarget(eid);
+            if(visitedVertices.find(nextVertice.x) == visitedVertices.end()){
+                bool loop_found = strongConnection( visitedVertices,
+                                                    verticesOnStack,
+                                                    graphRepresentation,
+                                                    graph,
+                                                    nextVertice);
+                if(loop_found){
+                    return true;
+                }
+            } else if (verticesOnStack.find(nextVertice.x) != verticesOnStack.end()){
+                return true;
+            }
+        } else {
+            continue;
+        }
+    }
+
+    verticesOnStack.erase(node.x);
+
+    return false;
+}
+
+bool TarjanLoopDetection(ChcDirectedGraph const & graph) {
+    if (graph.getVertices().size() < 3) { return false; }
+    auto graphRepresentation = AdjacencyListsGraphRepresentation::from(graph);
+    auto vertices = reversePostOrder(graph, graphRepresentation);
+    std::set<int> visitedVertices;
+    std::set<int> verticesOnStack;
+
+
+    for(uint i = 1; i < vertices.size()-1; i++){
+        if(visitedVertices.find(vertices[i].x) == visitedVertices.end()){
+            bool loop_detected = strongConnection(visitedVertices,
+                                                  verticesOnStack,
+                                                  graphRepresentation,
+                                                  graph,
+                                                  vertices[i]);
+            if (loop_detected) return true;
+        }
+    }
+    return false;
+}
+
 bool isTransitionSystemChain(ChcDirectedGraph const & graph) {
     if (graph.getVertices().size() < 3) { return false; }
     auto graphRepresentation = AdjacencyListsGraphRepresentation::from(graph);
     auto vertices = reversePostOrder(graph, graphRepresentation);
     assert(graph.getEntry() == vertices[0]);
     assert(graph.getExit() == vertices.back());
-    if (graphRepresentation.getOutgoingEdgesFor(vertices[0]).size() != 1
-        or graphRepresentation.getIncomingEdgesFor(vertices.back()).size() != 1) {
-        return false;
-    }
+    bool hasLoop = TarjanLoopDetection(graph);
+    if (hasLoop) {return false;}
     for (unsigned i = 1; i < vertices.size() - 1; ++i) {
         auto current = vertices[i];
         auto const & outEdges = graphRepresentation.getOutgoingEdgesFor(current);
-        if (outEdges.size() != 2) { return false; }
         bool hasSelfLoop = false;
         bool hasEdgeToNext = false;
         for (EId eid : outEdges) {
             hasSelfLoop |= graph.getTarget(eid) == current;
-            hasEdgeToNext |= graph.getTarget(eid) == vertices[i+1];
+            hasEdgeToNext |= graph.getTarget(eid) != current;
         }
         if (not (hasSelfLoop and hasEdgeToNext)) { return false; }
     }
--- a/src/TransformationUtils.cc
+++ b/src/TransformationUtils.cc
@@ -85,37 +85,36 @@ std::unique_ptr<TransitionSystem> toTransitionSystem(ChcDirectedGraph const & gr
     return ts;
 }
 
-bool strongConnection(std::set<int> & visitedVertices, std::set<int> & verticesOnStack,
-                      AdjacencyListsGraphRepresentation & graphRepresentation, ChcDirectedGraph const & graph,
+bool strongConnection(std::unordered_set<int> & visitedVertices, std::unordered_set<int> & verticesOnStack,
+                      AdjacencyListsGraphRepresentation const & graphRepresentation, ChcDirectedGraph const & graph,
                       SymRef node) {
     visitedVertices.insert(node.x);
     verticesOnStack.insert(node.x);
     auto const & outEdges = graphRepresentation.getOutgoingEdgesFor(node);
+    if (size(outEdges) <= 1) { return false; }
 
     for (EId eid : outEdges) {
-        if (size(outEdges) <= 1) { return false; }
         if (graph.getTarget(eid) != node) {
-            auto nextVertice = graph.getTarget(eid);
-            if (visitedVertices.find(nextVertice.x) == visitedVertices.end()) {
-                bool loop_found =
-                    strongConnection(visitedVertices, verticesOnStack, graphRepresentation, graph, nextVertice);
-                if (loop_found) { return true; }
-            } else if (verticesOnStack.find(nextVertice.x) != verticesOnStack.end()) {
+            auto nextVertex = graph.getTarget(eid);
+            if (visitedVertices.find(nextVertex.x) == visitedVertices.end()) {
+                bool loopFound =
+                    strongConnection(visitedVertices, verticesOnStack, graphRepresentation, graph, nextVertex);
+                if (loopFound) { return true; }
+            } else if (verticesOnStack.find(nextVertex.x) != verticesOnStack.end()) {
                 return true;
             }
         }
     }
 
     verticesOnStack.erase(node.x);
-
     return false;
 }
 
 bool TarjanLoopDetection(ChcDirectedGraph const & graph) {
     auto graphRepresentation = AdjacencyListsGraphRepresentation::from(graph);
     auto vertices = reversePostOrder(graph, graphRepresentation);
-    std::set<int> visitedVertices;
-    std::set<int> verticesOnStack;
+    std::unordered_set<int> visitedVertices;
+    std::unordered_set<int> verticesOnStack;
 
     for (uint i = 1; i < vertices.size() - 1; i++) {
         if (visitedVertices.find(vertices[i].x) == visitedVertices.end()) {
--- a/src/TransformationUtils.cc
+++ b/src/TransformationUtils.cc
@@ -19,19 +19,18 @@ bool isTransitionSystem(ChcDirectedGraph const & graph) {
     if (begOutEdges.size() != 1 || graph.getTarget(begOutEdges[0]) != loop) { return false; }
     auto const & loopOutEdges = graphRepresentation.getOutgoingEdgesFor(loop);
     if (loopOutEdges.size() != 2) { return false; }
-    bool selfLoop = std::find_if(loopOutEdges.begin(), loopOutEdges.end(),
-                                 [&graph, loop](EId eid) { return graph.getTarget(eid) == loop; }) !=
-                    loopOutEdges.end();
+    bool selfLoop = std::find_if(loopOutEdges.begin(), loopOutEdges.end(), [&graph, loop](EId eid) {
+                        return graph.getTarget(eid) == loop;
+                    }) != loopOutEdges.end();
     if (not selfLoop) { return false; }
     bool edgeToEnd = std::find_if(loopOutEdges.begin(), loopOutEdges.end(),
-                                  [&graph, end](EId eid) { return graph.getTarget(eid) == end; }) !=
-                     loopOutEdges.end();
+                                  [&graph, end](EId eid) { return graph.getTarget(eid) == end; }) != loopOutEdges.end();
     if (not edgeToEnd) { return false; }
     if (not graphRepresentation.getOutgoingEdgesFor(end).empty()) { return false; }
     return true;
 }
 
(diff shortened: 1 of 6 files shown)


src/engine/TPA.cc
```diff
@@ -1346,8 +1354,21 @@ class TransitionSystemNetworkManager {
...
+    void updateRestrictions(SymRef node);
 };
 
+void TransitionSystemNetworkManager::addRestrictions(SymRef node, PTRef fla) {
+    getNode(node).accumulatedRestrictions = logic.mkAnd(fla, getNode(node).accumulatedRestrictions);
+}
+
+void TransitionSystemNetworkManager::updateRestrictions(SymRef node) {
+    getNode(node).solver->updateQueryStates(getNode(node).accumulatedRestrictions);
+    getNode(node).accumulatedRestrictions = PTRef();
```
blishko, line 1369:
```suggestion
    getNode(node).accumulatedRestrictions = logic.getTerm_true();
```

src/engine/TPA.cc
```diff
@@ -1403,74 +1425,77 @@ VerificationResult TransitionSystemNetworkManager::solve() && {
...
-                continue; // Information has been propagated to the next node, continue querying that node
-
-            } else { // Edge cannot propagate forward
-                getNode(current).trulyReached = PTRef_Undef;
-//                std::cout << "Edge cannot propagate forward from: " << logic.pp(edgeExplanation) << std::endl;
-                getNode(current).solver->updateQueryStates(logic.mkNot(edgeExplanation));
-                continue; // Repeat the query for the same TS with stronger query
+            while(getNode(current).blocked_children < getNode(current).children.size()){
+                auto nextEdge = (*getOutgoingEdges(current))[getNode(current).blocked_children];
+                getNode(current).trulyReached = explanation;
```
blishko, line 1441: Why is this inside the loop?

src/engine/TPA.cc
```diff
@@ -1403,74 +1425,77 @@ VerificationResult TransitionSystemNetworkManager::solve() && {
...
+                getNode(current).blocked_children ++;
+                if (reachable(edgeRes)) { // Edge propagates forward
+                    auto next = graph.getTarget(nextEdge);
+                    networkMap.at(next).solver->resetInitialStates(edgeExplanation);
+                    activePath.push(nextEdge);
+                    current = next;
+                    break; // Information has been propagated to the next node, switch to the new node
+
+                } else { // Edge cannot propagate forward
+                    getNode(current).trulyReached = PTRef_Undef;
```
blishko, line 1453: This seems premature.
Truly reached states should be reset only after *all* outgoing edges has been blocked for them, not after one edge has been blocked.

src/engine/TPA.cc
```diff
@@ -1403,74 +1425,77 @@ VerificationResult TransitionSystemNetworkManager::solve() && {
...
         if (current == getChainEnd()) {
             break;
         }
-        EId next = getOutgoingEdge(current);
-        errorPath.push_back(next);
-        current = graph.getTarget(next);
+
     }
-    errorPath.push_back(getNode(current).next);
+    errorPath.push_back(getNode(current).children[0]);
```
blishko, line 1498: Error path should be built based on the activePath, not just taking the first child.

blishko (changes requested):
It is on the right path, but I think the extension from chain to DAG still needs work.

src/engine/TPA.cc
BritikovKI: It is built based on the active path, that's just the connection to the last element of the chain. At this stage `current == getChainEnd()`

src/engine/TPA.cc
```diff
@@ -1403,74 +1425,77 @@ VerificationResult TransitionSystemNetworkManager::solve() && {
     initNetwork();
 
     auto current = getChainStart();
```
blishko, line 1427: Note that now there does not have to be any `chain start`, because there is no chain anymore. In fact, now the exploration should start by traversing all the edges from `Entry` point of the graph.

src/TransformationUtils.cc
```diff
@@ -82,25 +83,86 @@ std::unique_ptr<TransitionSystem> toTransitionSystem(ChcDirectedGraph const & gr
     return ts;
 }
 
+//void strongconnect (SymRef v){
+//
+//}
```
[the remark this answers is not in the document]
BritikovKI: Fixed

src/TransformationUtils.cc
```diff
@@ -82,25 +83,86 @@ std::unique_ptr<TransitionSystem> toTransitionSystem(ChcDirectedGraph const & gr
...
+                                                    graphRepresentation,
+                                                    graph,
+                                                    nextVertice);
+                if(loop_found){
+                    return true;
+                }
+            } else if (verticesOnStack.find(nextVertice.x) != verticesOnStack.end()){
+                return true;
+            }
+        } else {
```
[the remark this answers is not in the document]
BritikovKI: Removed

src/TransformationUtils.cc
```diff
@@ -82,25 +83,86 @@ std::unique_ptr<TransitionSystem> toTransitionSystem(ChcDirectedGraph const & gr
...
+        }
+        if(graph.getTarget(eid) != node){
+            auto nextVertice = graph.getTarget(eid);
+            if(visitedVertices.find(nextVertice.x) == visitedVertices.end()){
+                bool loop_found = strongConnection( visitedVertices,
+                                                    verticesOnStack,
+                                                    graphRepresentation,
+                                                    graph,
+                                                    nextVertice);
+                if(loop_found){
```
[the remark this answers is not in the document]
BritikovKI: Should be fixed

src/TransformationUtils.cc
```diff
@@ -82,25 +83,86 @@ std::unique_ptr<TransitionSystem> toTransitionSystem(ChcDirectedGraph const & gr
...
+                                                  graphRepresentation,
+                                                  graph,
+                                                  vertices[i]);
+            if (loop_detected) return true;
+        }
+    }
+    return false;
+}
+
 bool isTransitionSystemChain(ChcDirectedGraph const & graph) {
```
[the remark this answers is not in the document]
BritikovKI: Renamed

src/engine/TPA.cc
```diff
@@ -1333,9 +1337,13 @@ class TransitionSystemNetworkManager {
 
     SymRef getChainEnd() const { return graph.getSource(adjacencyRepresentation.getIncomingEdgesFor(graph.getExit())[0]); }
 
-    EId getIncomingEdge(SymRef vid) const { return getNode(vid).previous; }
+    EId getIncomingEdge(SymRef vid, int id) const { return getNode(vid).parents[id]; }
+
+    const vec<EId>* getIncomingEdges(SymRef vid) const { return &getNode(vid).parents; }
```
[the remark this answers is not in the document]
BritikovKI: Changed

src/engine/TPA.cc
BritikovKI: Fixed

src/engine/TPA.cc
BritikovKI: Moved outside

src/engine/TPA.cc
BritikovKI: Moved to the state when everything is blcoked

blishko (changes requested):
Small comments

src/engine/TPA.cc
```diff
@@ -1329,13 +1333,29 @@ class TransitionSystemNetworkManager {
...
+        }
+        return graphStarts;
+    }
 
-    SymRef getChainEnd() const { return graph.getSource(adjacencyRepresentation.getIncomingEdgesFor(graph.getExit())[0]); }
+    vec<SymRef> getGraphEnds() const {
+        vec<SymRef> graphEnds;
+        for (uint i = 0; i < adjacencyRepresentation.getOutgoingEdgesFor(graph.getEntry()).size(); i++) {
+            graphEnds.push(graph.getSource(adjacencyRepresentation.getIncomingEdgesFor(graph.getExit())[i]));
+        }
```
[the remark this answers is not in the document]
BritikovKI: That was the reason for the failing test

test/test_TPA.cc
```diff
@@ -165,6 +165,47 @@ TEST_F(TPATest, test_TPA_chain_of_two_unsafe) {
...
+                                          ChcBody{{logic->mkEq(x, zero)}, {}}
+                                  },
+                                  { // S2(x) & x' = x + 1 => S2(x')
+                                          ChcHead{UninterpretedPredicate{predS2Next}},
+                                          ChcBody{{logic->mkEq(xp, logic->mkPlus(x, logic->mkIntConst(FastRational(1))))},
+                                                  {UninterpretedPredicate{predS2Current}}}
+                                  },
+                                  { // S1(x) & x' = x + 1 => S1(x')
+                                          ChcHead{UninterpretedPredicate{predS1Next}},
+                                          ChcBody{{logic->mkEq(xp, logic->mkPlus(x, logic->mkIntConst(FastRational(1))))},
```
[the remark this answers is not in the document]
BritikovKI: Fixed

src/engine/TPA.cc
```diff
@@ -1379,98 +1413,115 @@ void TransitionSystemNetworkManager::initNetwork() {
...
-//                std::cout << "Propagating forward query to " << next.id << ": " << logic.pp(edgeExplanation) << std::endl;
-                continue; // Information has been propagated to the next node, continue querying that node
-
-            } else { // Edge cannot propagate forward
+    auto graphStarts = getGraphStarts();
+    auto graphEnds = getGraphEnds();
+    bool safe;
+    for(auto init: graphStarts){
+        auto current = init;
+        activePath.shrink(activePath.size());
```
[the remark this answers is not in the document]
BritikovKI: Fixed

src/engine/TPA.cc
```diff
@@ -1379,98 +1413,115 @@ void TransitionSystemNetworkManager::initNetwork() {
...
+            if (reachable(res)) {
+                getNode(current).trulyReached = explanation;
+                for(auto end: graphEnds){
+                    if (current == end) {
+                        return {VerificationAnswer::UNSAFE, computeInvalidityWitness()};
+                    }
+                }
+                EId nextEdge;
+                while (getNode(current).blocked_children < getNode(current).children.size()) {
+                    nextEdge = (getOutgoingEdges(current))[getNode(current).blocked_children];
```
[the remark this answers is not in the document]
BritikovKI: Fixed

src/engine/TPA.cc
```diff
@@ -1305,10 +1305,14 @@ class TransitionSystemNetworkManager {
         std::unique_ptr<TPABase> solver {nullptr};
         PTRef trulyReached {PTRef_Undef};
         PTRef trulySafe {PTRef_Undef};
-        EId next;
-        EId previous;
+        PTRef accumulatedRestrictions;
+        vec<EId> children;
+        int blocked_children;
+        vec<EId> parents;
```
blishko, line 1311: I believe you do not need to store the parents. We are backtracking using the active path and that is enough.

src/engine/TPA.cc
```diff
@@ -1329,13 +1333,29 @@ class TransitionSystemNetworkManager {
...
+            graphEnds.push(graph.getSource(adjacencyRepresentation.getIncomingEdgesFor(graph.getExit())[i]));
+        }
+        return graphEnds;
+    }
 
-    EId getIncomingEdge(SymRef vid) const { return getNode(vid).previous; }
+    EId getIncomingEdge(SymRef vid, int id) const { return getNode(vid).parents[id]; }
 
-    EId getOutgoingEdge(SymRef vid) const { return getNode(vid).next; }
+    const vec<EId>& getIncomingEdges(SymRef vid) const { return getNode(vid).parents; }
```
blishko, line 1354: These are not needed.

src/engine/TPA.cc
```diff
@@ -1379,98 +1413,115 @@ void TransitionSystemNetworkManager::initNetwork() {
...
-                networkMap.at(next).solver->resetInitialStates(edgeExplanation);
-                current = next;
-//                std::cout << "Propagating forward query to " << next.id << ": " << logic.pp(edgeExplanation) << std::endl;
-                continue; // Information has been propagated to the next node, continue querying that node
-
-            } else { // Edge cannot propagate forward
+    auto graphStarts = getGraphStarts();
+    auto graphEnds = getGraphEnds();
+    bool safe;
+    for(auto init: graphStarts){
```
blishko, line 1441: Instead of iterating over graphStarts, you can just iterate over outgoing edges from CHCGraph entry.

src/engine/TPA.cc
```diff
@@ -1379,98 +1413,115 @@ void TransitionSystemNetworkManager::initNetwork() {
...
-                continue; // Blocked states have been propagated to the previous node, repeat the query there.
+            auto [res,explanation] = queryTransitionSystem(networkMap.at(current));
+            if (reachable(res)) {
+                getNode(current).trulyReached = explanation;
+                for(auto end: graphEnds){
+                    if (current == end) {
+                        return {VerificationAnswer::UNSAFE, computeInvalidityWitness()};
+                    }
+                }
+                EId nextEdge;
```
blishko, line 1460: This is used only inside the loop, no? No need to declare it outside the loop.

src/engine/TPA.cc
```diff
@@ -1379,98 +1413,115 @@ void TransitionSystemNetworkManager::initNetwork() {
...
-            } else { // Cannot continue from currently computed truly reached states
-                previousNode.trulyReached = PTRef_Undef;
-//                std::cout << "Strenghtening source TS with: " << logic.pp(edgeExplanation) << std::endl;
-                previousNode.solver->updateQueryStates(logic.mkNot(edgeExplanation));
-                current = graph.getSource(previousEdge);
-                continue; // Blocked states have been propagated to the previous node, repeat the query there.
+            auto [res,explanation] = queryTransitionSystem(networkMap.at(current));
+            if (reachable(res)) {
+                getNode(current).trulyReached = explanation;
+                for(auto end: graphEnds){
```
blishko, line 1455:
```suggestion
                for (auto end : graphEnds) {
```

blishko (commented):
Just a bunch of small comments.
I have a few other things that we can discuss on a call.
We can start testing this on CHC benchmarks. I need to think about the blocking of the edges.

Also, please set you formatting rules for this project to match its code style.

blishko:
I got convinced that the current version should work (which we should now test on CHC benchmarks), but I also got convinced that we need to shuffle the algorithm a bit, because now it does one thing in two places and it also can repeatedly compute the same reached states in one transition system for each of its outgoing edges.

src/engine/TPA.cc
BritikovKI: I think it saves some time for the performance, because it doesn't need to remove and create it each iteration

src/engine/TPA.cc
BritikovKI: The thing is, it will allow to possibly update our implementation, for example to parse the graph not from the beginning to the end, but in the reverse order. I can remove it of course, but it is a placer for a possible system updates, wdyt?

src/engine/TPA.cc
BritikovKI: Same thing as comment higher

src/engine/TPA.cc
```diff
@@ -609,7 +609,8 @@ TPASplit::QueryResult TPASplit::reachabilityQueryExact(PTRef from, PTRef to, uns
 //                std::cout << std::endl;
                 TRACE(3, "Learning " << itp.x)
                 TRACE(4, "Learning " << logic.pp(itp))
-                assert(itp != logic.getTerm_true());
+                // If itp == logic.getTerm_true, then the error states were trivially unreachable
+                if (itp == logic.getTerm_true()) { assert(power == 0); }
```
blishko, line 613: Here, I would keep the original assert. I believe we should not get here when the system is trivially safe.

src/TransformationUtils.cc
```diff
@@ -58,51 +58,92 @@ std::unique_ptr<TransitionSystem> toTransitionSystem(ChcDirectedGraph const & gr
...
     });
     assert(init != PTRef_Undef && transitionRelation != PTRef_Undef && bad != PTRef_Undef);
-    auto ts = std::unique_ptr<TransitionSystem>(new TransitionSystem(logic, std::move(systemType), init, transitionRelation, bad));
+    auto ts = std::unique_ptr<TransitionSystem>(
+        new TransitionSystem(logic, std::move(systemType), init, transitionRelation, bad));
     return ts;
 }
 
-bool isTransitionSystemChain(ChcDirectedGraph const & graph) {
+bool strongConnection(std::set<int> & visitedVertices, std::set<int> & verticesOnStack,
```
blishko, line 88: These could be `unordered_set`.

src/TransformationUtils.cc
```diff
@@ -58,51 +58,92 @@ std::unique_ptr<TransitionSystem> toTransitionSystem(ChcDirectedGraph const & gr
...
+                      AdjacencyListsGraphRepresentation & graphRepresentation, ChcDirectedGraph const & graph,
```
blishko, line 89:
```suggestion
                      AdjacencyListsGraphRepresentation const & graphRepresentation, ChcDirectedGraph const & graph,
```

src/TransformationUtils.cc
```diff
@@ -58,51 +58,92 @@ std::unique_ptr<TransitionSystem> toTransitionSystem(ChcDirectedGraph const & gr
...
+                      SymRef node) {
+    visitedVertices.insert(node.x);
+    verticesOnStack.insert(node.x);
+    auto const & outEdges = graphRepresentation.getOutgoingEdgesFor(node);
+
+    for (EId eid : outEdges) {
+        if (size(outEdges) <= 1) { return false; }
```
blishko, line 96: Is there an implicit assumption that if there is exactly one out-edge, it is a self-loop?

src/TransformationUtils.cc
```diff
@@ -58,51 +58,92 @@ std::unique_ptr<TransitionSystem> toTransitionSystem(ChcDirectedGraph const & gr
...
+        if (graph.getTarget(eid) != node) {
+            auto nextVertice = graph.getTarget(eid);
```
blishko, line 98:
```suggestion
            auto nextVertex = graph.getTarget(eid);
```

src/TransformationUtils.cc
```diff
@@ -58,51 +58,92 @@ std::unique_ptr<TransitionSystem> toTransitionSystem(ChcDirectedGraph const & gr
...
+            if (visitedVertices.find(nextVertice.x) == visitedVertices.end()) {
+                bool loop_found =
```
blishko, line 100:
```suggestion
                bool loopFound =
```

src/TransformationUtils.cc
```diff
@@ -58,51 +58,92 @@ std::unique_ptr<TransitionSystem> toTransitionSystem(ChcDirectedGraph const & gr
...
+                    strongConnection(visitedVertices, verticesOnStack, graphRepresentation, graph, nextVertice);
+                if (loop_found) { return true; }
+            } else if (verticesOnStack.find(nextVertice.x) != verticesOnStack.end()) {
+                return true;
+            }
+        }
+    }
+
+    verticesOnStack.erase(node.x);
+
```
blishko, line 110: Unnecessary empty line?

src/TransformationUtils.cc
blishko: Actually, why is this inside the loop?

src/engine/TPA.cc
```diff
@@ -6,16 +6,17 @@
...
 #include "QuantifierElimination.h"
+#include "TransformationUtils.h"
+#include "TransitionSystem.h"
 #include "graph/GraphTransformations.h"
 #include "transformers/BasicTransformationPipelines.h"
 
 #define TRACE_LEVEL 0
 
-#define TRACE(l,m) if (TRACE_LEVEL >= l) {std::cout << m << std::endl; }
+#define TRACE(l, m)                                                                                                    \
```
blishko, line 18: I would update clang-format config to not do this change.

src/engine/TPA.cc
```diff
@@ -562,10 +550,11 @@ TPASplit::QueryResult TPASplit::reachabilityQueryExact(PTRef from, PTRef to, uns
                     return result;
                 }
                 // Create the three states corresponding to current, next and next-next variables from the query
-//              PTRef modelMidpoint = getNextVersion(extractStateFromModel(getStateVars(1), *model), -1);
+                //              PTRef modelMidpoint = getNextVersion(extractStateFromModel(getStateVars(1), *model),
```
blishko, line 553: This space after start of comment is also weird...
Something to update in clang-format config.

src/engine/TPA.cc
```diff
@@ -1063,8 +1056,7 @@ bool TPASplit::checkExactFixedPoint(unsigned short power) {
 
 bool TPABase::verifyKinductiveInvariant(PTRef fla, unsigned long k) const {
     constexpr int trace_level = 1;
-    TRACE(trace_level, "Verifying k-inductive invariant for k = " << k)
-    {   // Inductive case:
+    TRACE(trace_level, "Verifying k-inductive invariant for k = " << k) { // Inductive case:
```
blishko, line 1059: This is incorrect change from clang-format. We need to make sure it does not do this.

src/engine/TPA.cc
```diff
@@ -1353,152 +1342,211 @@ class TransitionSystemNetworkManager {
     NetworkNode & getNode(SymRef vid) { return networkMap.at(vid); }
     NetworkNode const & getNode(SymRef vid) const { return networkMap.at(vid); }
 
-    SymRef getChainStart() const { return graph.getTarget(adjacencyRepresentation.getOutgoingEdgesFor(graph.getEntry())[0]); }
+    vec<SymRef> getGraphStarts() const {
+        vec<SymRef> graphStarts;
+        for (uint i = 0; i < adjacencyRepresentation.getOutgoingEdgesFor(graph.getEntry()).size(); i++) {
+            graphStarts.push(graph.getTarget(adjacencyRepresentation.getOutgoingEdgesFor(graph.getEntry())[i]));
+        }
```
blishko, line 1349: Store return value for `adjacencyRepresentation.getOutgoingEdgesFor` before the loop. Use range-based for loop afterwards.

src/engine/TPA.cc
```diff
@@ -1353,152 +1342,211 @@ class TransitionSystemNetworkManager {
...
+        return graphStarts;
+    }
 
-    SymRef getChainEnd() const { return graph.getSource(adjacencyRepresentation.getIncomingEdgesFor(graph.getExit())[0]); }
+    vec<SymRef> getGraphEnds() const {
+        vec<SymRef> graphEnds;
+        for (uint i = 0; i < adjacencyRepresentation.getIncomingEdgesFor(graph.getExit()).size(); i++) {
```
blishko, line 1355: Same as above.

src/TransformationUtils.cc
BritikovKI: Yup, I basically expect that if there's just one outgoing edge, then there is no loop for this specific node, bc it means that we just have a self-loop here.
I'll move it outside the loop

BritikovKI: Fixed

src/TransformationUtils.cc
BritikovKI: Fixed

src/TransformationUtils.cc
BritikovKI: Fixed

src/engine/TPA.cc
BritikovKI: Fixed

src/engine/TPA.cc
BritikovKI: Fixed

blishko (approved):
LGTM!

PR #19, opened 2023-03-04
TPA: Handle properly special cases of trivial transition systems

blishko:
If TPA was queried with a transition system that has empty bad states, it ran into an assertion that the interpolant is not logical true. This assertion is, in general, useful to detect incorrect behaviour of the algorithm, so we would like to keep it, even though in this special case of `query == false` such interpolant is correct. Instead, we check for these special cases beforehand.

For now we only make syntactic check. It is possible we should do proper SMT check on the initial and query states.

### DIFF
--- a/src/engine/TPA.cc
+++ b/src/engine/TPA.cc
@@ -392,6 +392,11 @@ VerificationAnswer TPABase::solveTransitionSystem(TransitionSystem & system) {
 }
 
 VerificationAnswer TPABase::solve() {
+    auto res = checkTrivialUnreachability();
+    assert(res != VerificationAnswer::UNSAFE);
+    if (res == VerificationAnswer::SAFE) {
+        return res;
+    }
     unsigned short power = 0;
     while (true) {
         auto res = checkPower(power);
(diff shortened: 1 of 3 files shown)


src/engine/TPA.cc
```diff
@@ -392,6 +392,11 @@ VerificationAnswer TPABase::solveTransitionSystem(TransitionSystem & system) {
 }
 
 VerificationAnswer TPABase::solve() {
+    auto res = checkTrivialUnreachability();
+    assert(res != VerificationAnswer::UNSAFE);
```
BritikovKI, line 396: Why even have this assert here?
It's unreachable in any case, because it is never returned from `checkTrivialUnreachability`

src/engine/TPA.cc
```diff
@@ -405,6 +410,25 @@ VerificationAnswer TPABase::solve() {
     }
 }
 
+VerificationAnswer TPABase::checkTrivialUnreachability() {
+    if (query == logic.getTerm_false()) {
+        // TODO: Check UNSAT with solver?
+        explanation.power = 0;
+        explanation.relationType = TPAType::LESS_THAN;
```
BritikovKI, line 417: Btw, just interesting, why this specific type of relation, can you pls briefly explain this parameters, bc I'm a little bit puzzled?

src/engine/TPA.cc
blishko: At the moment, yes. It's just a check for some future person that would try to change something in the `checkTrivialUnreachability` method.

src/engine/TPA.cc
blishko: `TPAType::LESS_THAN` should be used so it works for both TPA modes.

The others are probably not necessarily, but should be set to reasonable values so that methods `getSafetyExplanation` and `getInductiveInvariant` work correctly.

Except `power`, which must be set to `0`.
The over-approximation on level 0 is `true` at this point, so fixed point type and invariant type can be anything, but these are a bit safer values. If `query` is `false`, then all transition formulas are invariants when restricted to `query`.
Similarly, if `init` is `false`, then all transition formulas are invariants when restricted to `init`.

BritikovKI (approved):
Lgtmt

PR #20, opened 2023-03-15
TPA: Add functionality for solving trivial systems

blishko:
This is a quick fix TPA to be able to handle instances with no loops.
Since TPA eliminates non-looping nodes, the trivial graph will have (0 or more) edges from entry to exit.

This should be probably generalized.
Nonrecursive CHC systems can be solved prior to passing the problem to the engines.
One way is to solve it as a single tree-interpolation problem.
An easier way (with less new code) is to apply non-loop-node elimination to obtain a trivial graph and let the backtranslation compute the witness.


src/engine/TPA.cc
```diff
@@ -46,7 +46,44 @@ VerificationResult TPAEngine::solve(ChcDirectedHyperGraph & graph) {
     return VerificationResult(VerificationAnswer::UNKNOWN);
 }
 
+bool TPAEngine::isTrivial(ChcDirectedGraph const & graph) {
+    return graph.getVertices().size() <= 2;
+}
+
+VerificationResult TPAEngine::solveTrivial(ChcDirectedGraph const & graph) {
+    // All edges should be between entry and exit, check if any of them has a satisfiable label
+    auto edgeIds = graph.getEdges();
```
BritikovKI, line 55:
```suggestion
    auto edgeIds = graph.getEdges();
    assert(edgeIds.size() <= 1);
```

BritikovKI (approved):
Lgtmt

PR #21, opened 2023-04-03
Multiprocessing

BritikovKI:
Added profiling option to the golem, which runs multiple processes based on the configuration


blishko:
OK!

PR #30, opened 2023-06-03
Spacer: Push lemmas in batches

blishko:
During the check for inductive invariant, we were previously trying to push lemmas to the next level one by one.
Now,  we attempt to push them in batches, gradually dropping those that cannot be pushed to the next level.

We use activation literals to track which lemmas we already showed cannot be pushed to disable them from further considerations. We basically simulate `check-sat-assuming`, which OpenSMT does not support yet, with `push` and `pop`.
The idea is the following: 
We maintain a record of all active lemmas, which could still be pushed.
At each iteration, we try to push all active lemmas. If this is not possible, we obtain a model, a counterexample-to-induction (CTI). We iterate over all active lemmas and disable those that are falsified by this model (we actually maintain the list of negation of lemmas, so we check if the negations are true in the model).
We update the activation literals to disable falsified lemmas from further consideration and repeat the whole process.
The process stops when we successfully push all remaining active lemmas or if we disable all lemmas.

This had overall very positive effect on both LIA-Lin and LIA-nonlin benchmarks.

### DIFF
--- a/src/engine/Spacer.cc
+++ b/src/engine/Spacer.cc
@@ -568,6 +568,8 @@ SpacerContext::InductiveCheckResult SpacerContext::isInductive(std::size_t maxLe
     return InductiveCheckResult{InductiveCheckAnswer::NOT_INDUCTIVE, 0};
 }
 
+/* This is the original tryPushComponents implementation that tries to push the lemmas one by one */
+#if 0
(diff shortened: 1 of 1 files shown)


blishko:
LIA-Lin:
comparison

LIA-nonlin:
comparison

BritikovKI (approved):
Lgtmt

PR #31, opened 2023-06-07
Transition subinvariant

BritikovKI:


src/engine/TPA.cc
```diff
@@ -982,16 +983,101 @@ bool TPASplit::verifyExactPower(unsigned short power) const {
     return res == s_False;
 }
 
+vec<PTRef> TPABase::houdiniCheck(PTRef invCandidates, PTRef transition, SafetyExplanation::FixedPointType allignment) {
```
blishko, line 986:
```suggestion
vec<PTRef> TPABase::houdiniCheck(PTRef invCandidates, PTRef transition, SafetyExplanation::FixedPointType alignment) {
```

src/engine/TPA.cc
```diff
@@ -982,16 +983,101 @@ bool TPASplit::verifyExactPower(unsigned short power) const {
...
+    SMTConfig config;
+    MainSolver solver(logic, config, "Fixed-point checker");
+    solver.push();
+    auto candidates = topLevelConjuncts(logic, invCandidates);
+    if (allignment == SafetyExplanation::FixedPointType::RIGHT) {
+        solver.insertFormula(getNextVersion(transition));
+    } else {
+        if (allignment == SafetyExplanation::FixedPointType::LEFT) { solver.insertFormula(transition); }
+    }
```
blishko, line 995:
```suggestion
    } else if (allignment == SafetyExplanation::FixedPointType::LEFT) { 
        solver.insertFormula(transition);
    }
```

src/engine/TPA.cc
```diff
@@ -982,16 +983,101 @@ bool TPASplit::verifyExactPower(unsigned short power) const {
...
+    //     shiftOnlyNextVars(currentLevelTransition);
+    while (candidates.size() > 128) {
+        for (int i = candidates.size() - 1; i >= 1; i -= 2) {
+            PTRef n_f = logic.mkAnd(candidates[i], candidates[i - 1]);
+            candidates.pop();
+            candidates.pop();
+            candidates.push(n_f);
+            if (candidates.size() <= 128) { break; }
+        }
+    }
```
blishko, line 1012: I think this is not doing what it was intended to do.
I tried to simulate what it is doing and I believe this is going to basically end up removing all elements after the cutoff (128), instead of combining them into a single conjunction.

src/engine/TPA.cc
```diff
@@ -982,16 +983,101 @@ bool TPASplit::verifyExactPower(unsigned short power) const {
...
+    //    invCandidates.append(conjuncts);
+    //    Atr(x, x') /\ tr(x', x'') => Atr(x, x'')
+    //    or
+    //    Atr(x', x'') /\  tr(x, x') => Atr(x, x'')
+    //    Push transition once and for all
+    //    While loop externally, because we may drop smth important
+    PTRef goal = shiftOnlyNextVars(logic.mkAnd(candidates));
```
blishko, line 1019: Compute `logic.mkAnd(candidates)` only once and use the resulted `PTRef` everywhere.
No need to compute the conjunction repeatedly.

src/engine/TPA.cc
```diff
@@ -982,16 +983,101 @@ bool TPASplit::verifyExactPower(unsigned short power) const {
...
+            if (allignment == SafetyExplanation::FixedPointType::RIGHT) {
+                solver.insertFormula(logic.mkAnd({logic.mkAnd(candidates), logic.mkNot(shiftOnlyNextVars(cand))}));
+            } else {
+                if (allignment == SafetyExplanation::FixedPointType::LEFT) {
+                    solver.insertFormula(
+                        logic.mkAnd({getNextVersion(logic.mkAnd(candidates)), logic.mkNot(shiftOnlyNextVars(cand))}));
+                }
+            }
+            if (solver.check() == s_True) {
+                Model model = *solver.getModel();
```
blishko, line 1042: This variable is never used. Why do you call `getModel` here?

src/engine/TPA.cc
```diff
@@ -982,16 +983,101 @@ bool TPASplit::verifyExactPower(unsigned short power) const {
...
+            solver.push();
+            if (allignment == SafetyExplanation::FixedPointType::RIGHT) {
+                solver.insertFormula(logic.mkAnd({logic.mkAnd(candidates), logic.mkNot(shiftOnlyNextVars(cand))}));
+            } else {
+                if (allignment == SafetyExplanation::FixedPointType::LEFT) {
+                    solver.insertFormula(
+                        logic.mkAnd({getNextVersion(logic.mkAnd(candidates)), logic.mkNot(shiftOnlyNextVars(cand))}));
+                }
+            }
+            if (solver.check() == s_True) {
```
blishko, line 1041: I think you do not need to do this checks like this. I think it would be better to just get the model from the big query and remove all the candidates whose `shiftOnlyNextVars` version is falsified in this model.
Then repeat the big query with a smaller set of candidates.

src/engine/TPA.cc
```diff
@@ -982,16 +983,101 @@ bool TPASplit::verifyExactPower(unsigned short power) const {
...
+    }
+    for (auto cand : candidates) {
+        if (allignment == SafetyExplanation::FixedPointType::RIGHT) {
+            if (std::find(rightInvariants.begin(), rightInvariants.end(), cand) != rightInvariants.end()) { continue; }
+            rightInvariants.push(cand);
+        } else {
+            if (std::find(leftInvariants.begin(), leftInvariants.end(), cand) != leftInvariants.end()) { continue; }
+            leftInvariants.push(cand);
+        }
+    }
```
blishko, line 1065: This should be better to do at the beginning, then you don't have to do SMT queries for this candidates at all.
Also, it would be more efficient to keep the invariants in a set as well as in a vector. So that the membership check would be much cheaper.

src/engine/TPA.cc
```diff
@@ -982,16 +983,101 @@ bool TPASplit::verifyExactPower(unsigned short power) const {
...
+    return candidates;
```
blishko, line 1066: I don't see the returned value used anywhere in callers.

src/engine/TPA.cc
```diff
@@ -1049,6 +1137,54 @@ bool TPABase::checkLessThanFixedPoint(unsigned short power) {
...
+            if (satres == s_False) {
+                if (verbose() > 0) {
+                    std::cout << "; Left fixed point detected in less-than relation on level " << i << " from " << power
+                              << std::endl;
+                    std::cout << "; Fixed point detected for "
+                              << (not restrictedInvariant ? "whole transition relation"
+                                                          : "transition relation restricted to bad")
+                              << std::endl;
+                }
+                explanation.invariantType = restrictedInvariant
```
blishko, line 1155: Here the invariant type is always unrestricted, I believe.

src/engine/TPA.cc
```diff
@@ -1049,6 +1137,54 @@ bool TPABase::checkLessThanFixedPoint(unsigned short power) {
...
+            bool restrictedInvariant = false;
+            if (satres == s_False) {
+                if (verbose() > 0) {
+                    std::cout << "; Left fixed point detected in less-than relation on level " << i << " from " << power
+                              << std::endl;
+                    std::cout << "; Fixed point detected for "
+                              << (not restrictedInvariant ? "whole transition relation"
+                                                          : "transition relation restricted to bad")
+                              << std::endl;
+                }
```
blishko, line 1154: Here these claims are not correct anymore.

src/engine/TPA.cc
```diff
@@ -1049,6 +1137,54 @@ bool TPABase::checkLessThanFixedPoint(unsigned short power) {
...
+                explanation.invariantType = restrictedInvariant
+                                                ? SafetyExplanation::TransitionInvariantType::RESTRICTED_TO_QUERY
+                                                : SafetyExplanation::TransitionInvariantType::UNRESTRICTED;
+                explanation.relationType = TPAType::LESS_THAN;
+                explanation.power = i;
```
blishko, line 1159: This does not make sense anymore, because the invariants can come from different levels.

src/engine/TPA.cc
```diff
@@ -1057,7 +1193,7 @@ bool TPASplit::checkExactFixedPoint(unsigned short power) {
     assert(verifyExactPower(power + 1));
     for (unsigned short i = 1; i <= power + 1; ++i) {
         PTRef currentLevelTransition = getExactPower(i);
-        PTRef currentTwoStep = logic.mkAnd(currentLevelTransition, getNextVersion(currentLevelTransition));
+        PTRef currentTwoStep = logic.mkAnd({currentLevelTransition, getNextVersion(currentLevelTransition)});
```
blishko, line 1196: No need for the curly brackets.

blishko (changes requested):
I think this still needs some modifications.

src/engine/TPA.cc
blishko: Please refactor this loop to a separate method.

src/engine/TPA.cc
BritikovKI: But I need to update the formula when I update set of candidates

blishko: That's true.
But you use the same conjunction for the goal and then when inserting into the solver. So it still makes sense to compute the conjunction only once and use it twice.

blishko (approved):
Looks good!
I'll squash the commits into a single commit

PR #32, opened 2023-06-10
Spacer: Check must reachability for all edges with single SMT query

blishko:
Instead of checking each edge separately, we can batch the checks for all edges into one SMT query.
If any edge's must summary can be used, we can figure out which edge it is by checking each summary against the discovered model.


BritikovKI (approved):
Lgtmt

PR #33, opened 2023-06-27
Engines: Refactor computation of witnesses for transition systems

blishko:
This refactoring prepares the code for transformation of general linear CHC system to a transition system. The plan is that there will be no CHC graph corresponding to a transition system, but we go to directly to our inner TS representation. For this reason, we refactor the methods so they do not require both representation of a transition system.

To support witness production, we add a simple wrapper for a witness for our inner TS representation.

### DIFF
--- a/src/Witnesses.cc
+++ b/src/Witnesses.cc
@@ -178,3 +178,17 @@ ValidityWitness::fromTransitionSystem(Logic & logic, ChcDirectedGraph const & gr
     ValidityWitness::definitions_t definitions{{unversionedPredicate, graphInvariant}};
     return ValidityWitness(std::move(definitions));
 }
+
+VerificationResult
+translateTransitionSystemResult(TransitionSystemVerificationResult result, ChcDirectedGraph const & graph, TransitionSystem const & transitionSystem) {
+    switch (result.answer) {
+        case VerificationAnswer::SAFE:
+            return {VerificationAnswer::SAFE, ValidityWitness::fromTransitionSystem(graph.getLogic(), graph, transitionSystem, std::get<PTRef>(result.witness))};
+        case VerificationAnswer::UNSAFE:
+            return {VerificationAnswer::UNSAFE, InvalidityWitness::fromTransitionSystem(graph, std::get<std::size_t>(result.witness))};
+        case VerificationAnswer::UNKNOWN:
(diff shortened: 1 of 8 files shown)


src/Witnesses.cc
```diff
@@ -177,4 +177,18 @@ ValidityWitness::fromTransitionSystem(Logic & logic, ChcDirectedGraph const & gr
...
+    switch (result.answer) {
+        case VerificationAnswer::SAFE:
+            return {VerificationAnswer::SAFE, ValidityWitness::fromTransitionSystem(graph.getLogic(), graph, transitionSystem, std::get<PTRef>(result.witness))};
+        case VerificationAnswer::UNSAFE:
+            return {VerificationAnswer::UNSAFE, InvalidityWitness::fromTransitionSystem(graph, std::get<std::size_t>(result.witness))};
+        case VerificationAnswer::UNKNOWN:
+            return VerificationResult(VerificationAnswer::UNKNOWN);
+    }
+    assert(false);
+    return VerificationResult(VerificationAnswer::UNKNOWN);
```
BritikovKI, line 193: Why have return here if you already have assert(false) higher?

src/engine/Bmc.cc
```diff
@@ -8,15 +8,20 @@
...
-        return solveTransitionSystem(*ts, system);
+VerificationResult BMC::solve(ChcDirectedGraph const & graph) {
+    if (isTransitionSystem(graph)) {
+        return solveTransitionSystem(graph);
     }
     return VerificationResult(VerificationAnswer::UNKNOWN);
 }
 
-VerificationResult BMC::solveTransitionSystem(TransitionSystem const & system, ChcDirectedGraph const & graph) {
+VerificationResult BMC::solveTransitionSystem(ChcDirectedGraph const & graph) {
```
BritikovKI, line 18: Why make 1 func into 2, can we just do line 19 and line 21 inside of the "solveTransitionSystemInternal", but it's Ok I think

It's just a function for 2 lines of code

BritikovKI (approved):
Lgtmt

blishko:
I need the separation of the concepts for the transformation that I want to implement.
Basically, I need a witness for a transition system without a `ChcDirectedGraph`, which is not possible in current implementation.

If the functions would be joined, then I would have to call the translation method on every return statement, here there is only one invocation.

PR #34, opened 2023-07-11
Implement tranformation from a general linear CHC system into a transition system

blishko:
This PR implements the translation from a general linear CHC system to a transition system.
The implementation follows the algorithm described in the paper `Horn2VMT: Translating Horn Reachability into Transition Systems`
https://www.osti.gov/servlets/purl/1783647

The idea is to use additional variable(s) to represent location in the graph induced by the CHC system and define a single transition relation with disjuncts representing each transition in the original system.

The solution to the transition system can be translated to a witness for the CHC system as follows:

1.  From a TS error trace we can identify which locations were visited and which edges were used for each step of the trace. (The implementation currently identifies only locations. If there exists multiple edges between the source and target location, we bail out. The proper way is to check which disjunct of the transition relation was responsible for the step).
2. For an invariant we computed a projected invariant for each location by enabling the location and disabling all others. However, this can still leave unrelated variables in the formula. It is not yet clear to me what approach is best to deal with these variables. Currently, the implementation tries to find equivalent formula only over allowed variables using simplification techniques.

The transformation is enabled for BMC and IMC engines. For KIND, this approach does not work at the moment, because the transformation yields a safety property that is never k-inductive on its own. (The property says that error location is never reached). I think that for KIND a different way how to extend the algorithm to linear CHC systems is more appropriate. The transformation could work for PD-KIND which automatically strengthens the property.


src/Validator.cc
```diff
@@ -7,6 +7,7 @@
 #include "Validator.h"
 
 Validator::Result Validator::validate(ChcDirectedHyperGraph const & graph, VerificationResult const & result) {
+    if (not result.hasWitness()) { return Validator::Result::NOT_VALIDATED; }
```
BritikovKI, line 10: That's for some of the cases, when we shrink graph to single TS?

BritikovKI: Same question about `noWitness` structure

src/engine/Common.h
```diff
@@ -0,0 +1,9 @@
+#ifndef GOLEM_COMMON_H
+#define GOLEM_COMMON_H
+
+#include "graph/ChcGraph.h"
+#include "Witnesses.h"
+
+VerificationResult solveTrivial(ChcDirectedGraph const & graph);
```
BritikovKI, line 7: Hmm, should it be a separate "engine" though? Maybe `utils`? Wdyt?

BritikovKI (commented):
Checking the main part also, will finalize tomorrow

src/engine/Common.h
blishko: Yes, this is basically `utils` for all engines. What do you suggest? Rename the file?

src/Validator.cc
blishko: There are other situations where we might not be able to produce the witness at the moment.

src/engine/Common.h
BritikovKI: Maybe create an "utils" folder in "engines" and reuse it? Wdyt?

blishko: I am not very keen on expanding the folder structure at the moment. Maybe if there are more things to put there in the future...

BritikovKI:
Btw, do you have the results for BMC or IMC with translation?
Just to check the correctness in comparison with Z3 jic

src/transformers/SingleLoopTransformation.cc
```diff
@@ -0,0 +1,286 @@
...
+    PTRef initialStates = [&]() -> PTRef {
+        vec<PTRef> negatedLocations;
+        negatedLocations.capacity(locationVars.size());
+        for (auto && entry : locationVars) {
+            negatedLocations.push(logic.mkNot(entry.second));
+        }
+        return logic.mkAnd(std::move(negatedLocations));
+    }();
+
+    PTRef badStates = locationVars.at(graph.getExit());
```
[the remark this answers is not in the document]
blishko: There is always just a single `Exit`. Similarly there is always just a single `Entry` in the graph.

src/transformers/SingleLoopTransformation.cc
```diff
@@ -0,0 +1,286 @@
...
+    Logic & logic = graph.getLogic();
+    TimeMachine timeMachine(logic);
+    auto vertices = graph.getVertices();
+    // MB: It is useful to have exit location, so we do not remove exit from the vertices
+    vertices.erase(std::remove_if(vertices.begin(), vertices.end(), [&](auto vertex) {
+                       return vertex == graph.getEntry();
+                   }), vertices.end());
+    LocationVarMap locationVars;
+    locationVars.reserve(vertices.size());
+    for (auto vertex : vertices) {
```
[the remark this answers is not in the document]
blishko: There is no location variable for entry point.
The exit node has a location variable because that becomes the new safety property, that this variable never becomes true. Maybe this translation could be optimized, but that is for another time.

src/transformers/SingleLoopTransformation.cc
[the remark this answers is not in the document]
BritikovKI: Oh yup, srry, got mixed, there can be multiple edges to the end, not multiple "ends"

BritikovKI (approved):
Lgtmt

src/transformers/SingleLoopTransformation.h
```diff
@@ -0,0 +1,97 @@
...
+        SymRef vertex;
+        uint32_t pos;
+
+        inline bool operator==(VarPosition other) const { return vertex == other.vertex and pos == other.pos; }
+    };
+    struct VarPositionHasher {
+        std::size_t operator()(VarPosition pos) const {
+            std::hash<std::size_t> hasher;
+            std::size_t res = 0;
+            res ^= hasher(pos.vertex.x) + 0x9e3779b9 + (res<<6) + (res>>2);
```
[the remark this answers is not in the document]
blishko: Fixed now to use a simpler hashing of two values.

PR #36, opened 2023-07-21
Fix issues with witnesses for IMC and some transformation passes

blishko:
The main change here is a fix to IMC to produce correct inductive invariant for the given transition system.
The problem was that when IMC detects a fixed point, the state formula is an inductive invariant, but it is not necessarily safe on its own. We only know that it cannot reach bad states in `k` steps where `k` is the current value of the parameter of the procedure `finiteRun`.
While this is sound, because it still guarantees that bad states are unreachable, the witness has to be different.
For now I only was able to prove that the fixed point conjoined with the negation of bad states is k-inductive.
(It would be good to try to improve on this to avoid k-induction).

While working on this, I discovered a bug in `BackTranslator` for `NodeEliminator` which was not tracking the necessary information correctly and this resulted in incorrect final proofs.

Additionally, this PR adds the backtranslator for `MultiEdgeMerger` (which merges multiple edges with the same source and target to a single edge). This enables us to use `MultiEdgeMerger` for the preprocessing for all engines. This enabled `IMC` to solve more benchmarks and compute witnesses correctly for all UNSAT cases which it was  able to solve.

### DIFF
--- a/src/Witnesses.h
+++ b/src/Witnesses.h
@@ -104,7 +104,17 @@ public:
                                                 TransitionSystem const & transitionSystem, PTRef invariant);
 };
 
-struct NoWitness {};
+class NoWitness {
+    std::string reason;
+public:
+    explicit NoWitness(std::string_view reason) : reason(reason) {}
+
+    NoWitness() = default;
+    NoWitness(NoWitness &&) = default;
+    NoWitness & operator=(NoWitness &&) = default;
+
+    [[nodiscard]] std::string_view getReason() const { return reason; }
+};
 
 enum class VerificationAnswer : char {SAFE, UNSAFE, UNKNOWN};
 
@@ -119,6 +129,9 @@ public:
     VerificationResult(VerificationAnswer answer, InvalidityWitness invalidityWitness)
         : answer(answer), witness(std::move(invalidityWitness)) {}
 
+    VerificationResult(VerificationAnswer answer, NoWitness noWitness)
+        : answer(answer), witness(std::move(noWitness)) {}
+
     explicit VerificationResult(VerificationAnswer answer) : answer(answer), witness(NoWitness()) {}
 
     [[nodiscard]] VerificationAnswer getAnswer() const { return answer; }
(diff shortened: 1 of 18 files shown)


src/transformers/SingleLoopTransformation.h
```diff
@@ -70,9 +70,11 @@ class SingleLoopTransformation {
         VerificationResult translate(TransitionSystemVerificationResult result);
 
     private:
-        InvalidityWitness translateErrorPath(std::size_t unrolling);
+        template<typename T> using ErrorOr = std::variant<NoWitness, T>;
```
BritikovKI, line 73: Name is a little bit weird
Maybe `EmptyOrWitness` or smth?

src/transformers/MultiEdgeMerger.cc
```diff
@@ -1,23 +1,82 @@
 /*
- * Copyright (c) 2022, Martin Blicha <martin.blicha@gmail.com>
+ * Copyright (c) 2022-2023, Martin Blicha <martin.blicha@gmail.com>
```
BritikovKI, line 2: Can you add this file and .h to the formatter pls?

src/Witnesses.h
```diff
@@ -72,6 +75,7 @@ class InvalidityWitness {
     void setDerivation(Derivation derivation_) { derivation = std::move(derivation_); }
 
     [[nodiscard]] Derivation const & getDerivation() const { return derivation; }
+    [[nodiscard]] Derivation & getDerivation() { return derivation; }
```
BritikovKI, line 78: Hmm, why 2 functions with different modifiers?

src/graph/ChcGraph.cc
```diff
@@ -410,11 +410,12 @@ ChcDirectedHyperGraph::VertexContractionResult ChcDirectedHyperGraph::contractVe
     std::transform(outgoingEdges.begin(), outgoingEdges.end(), std::back_inserter(result.outgoing), [this](EId eid) {
         return this->getEdge(eid);
     });
-    std::size_t incomingIndex = 0;
-    for (EId incomingId : incomingEdges) {
+    for (std::size_t incomingIndex = 0; incomingIndex < incomingEdges.size(); ++incomingIndex) {
```
BritikovKI, line 413: Why change the loop conditions?

src/Witnesses.h
blishko: Because I needed also the non-`const` version. That is a pretty common pattern.

src/graph/ChcGraph.cc
blishko: To make sure I don't forget to increment the index, which was the bug in the previous version of the code.

src/transformers/SingleLoopTransformation.h
blishko: `ErrorOr<Witness>` captures the semantic I wanted to express. I just did not want to write the whole templated class for that.
The functions where it is used return `Validity/InvalidityWitness` under normal circumstances, but in some cases they may not be able to compute the witness, which is the `Error` result.

src/transformers/MultiEdgeMerger.cc
blishko: Done

src/Witnesses.h
BritikovKI: Ok, gotcha

BritikovKI (approved):
Lgtmt

PR #37, opened 2023-08-02
TPA: Fix computation of explanations

blishko:
In the current version method TPABase::getSafetyExplanation does not work correctly if the safety has been proven using (left or right) transition invariants collected in the Houdini check.

The problem is that the method in the current version still relies on the information stored in explanation.power, but this information is not (and cannot) be filled in the case mentioned above.

The proposed solution is to store directly the transition invariant in the explanation, instead of implicitly relying on the assumption that the invariant is one of the abstract transitions.

### DIFF
--- a/src/engine/TPA.cc
+++ b/src/engine/TPA.cc
@@ -69,7 +69,7 @@ VerificationResult TPAEngine::solve(const ChcDirectedGraph & graph) {
             case VerificationAnswer::SAFE: {
                 PTRef inductiveInvariant = solver->getInductiveInvariant();
                 if (inductiveInvariant == PTRef_Undef) { return VerificationResult(res); }
-                //                std::cout << "TS invariant: " << logic.printTerm(inductiveInvariant) << std::endl;
+                // std::cout << "TS invariant: " << logic.printTerm(inductiveInvariant) << std::endl;
                 return VerificationResult(res, computeValidityWitness(graph, *ts, inductiveInvariant));
             }
             case VerificationAnswer::UNKNOWN:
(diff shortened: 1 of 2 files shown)


src/engine/TPA.cc
```diff
@@ -1769,12 +1782,10 @@ PTRef TPABase::getSafetyExplanation() const {
         // TODO: compute the safe inductive invariant and return negation of that?
         return init;
     }
-    auto power = explanation.power;
-    PTRef transitionInvariant = explanation.relationType == TPAType::LESS_THAN
-                                    ? getPower(power, explanation.relationType)
-                                    : logic.mkOr(shiftOnlyNextVars(getPower(power, TPAType::LESS_THAN)),
-                                                 logic.mkAnd(getPower(power, TPAType::LESS_THAN),
-                                                             getNextVersion(getPower(power, TPAType::EQUALS))));
```
blishko, line 1777: Here was the problem before. This computation was not correct if we got here after with concluded that the system is safe because of (left or right) invariants computed from the Houdini check.

BritikovKI: Hmm, Ok, you basically reworked the data structure as I can see...

BritikovKI (approved):
Lgtmt

src/engine/TPA.cc
blishko: Yes, instead of remembering which abstract relation is the transition invariant, we store the transition invariant directly.
Instead of `power`, which indicated which abstract relation should be used, we now remember the value `k` for which the transition invariant is k-inductive. Or rather, we store the exponent of the power of 2. That is, we store `exp` so that the invariant is `2^exp`-inductive.

PR #38, opened 2023-08-04
Everywhere: Simplify typical use of SMT solver

blishko:
The goal is to provide a simple abstraction over OpenSMT's solver so that we do not have to repeat unnecessary complicated setup of OpenSMT's SMTConfig and MainSolver. It would also better contain future changes in OpenSMT.

In the future, this abstraction can be extended to completly isolate OpenSMT's solver in a single place.

### DIFF
--- a/src/utils/SmtSolver.cc
+++ b/src/utils/SmtSolver.cc
@@ -1,0 +1,21 @@
+/*
+ * Copyright (c) 2023, Martin Blicha <martin.blicha@gmail.com>
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+#include "SmtSolver.h"
+
+SMTSolver::SMTSolver(Logic & logic, WitnessProduction setup) {
+    bool produceModel = setup == WitnessProduction::ONLY_MODEL || setup == WitnessProduction::MODEL_AND_INTERPOLANTS;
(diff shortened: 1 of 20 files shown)


src/utils/SmtSolver.cc
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright (c) 2023, Martin Blicha <martin.blicha@gmail.com>
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+#include "SmtSolver.h"
+
+SMTSolver::SMTSolver(Logic & logic, WitnessProduction setup) {
```
BritikovKI, line 9: Maybe also make config as a part of constructor?  What if we need engine-specific configuration?

blishko: You can get the config with `getConfig` method and set what you need there.
Requiring `SMTConfig` in the constructor would defeat the purpose of the proposed change.

Later, I imagine we identify what we require and have methods for that, so that we can hide `MainSolver` and `SMTConfig` completely.

BritikovKI: Gotcha

BritikovKI (approved):
Lgtmt

PR #44, opened 2023-09-20
Alethe proof generation

M4mbo:


src/Options.cc
```diff
@@ -76,16 +81,18 @@ Options CommandLineParser::parse(int argc, char ** argv) {
             {Options::VALIDATE_RESULT.c_str(), no_argument, &validate, 1},
             {Options::PRINT_WITNESS.c_str(), no_argument, &printWitness, 1},
             {Options::COMPUTE_WITNESS.c_str(), optional_argument, &computeWitness, 1},
-            {Options::LRA_ITP_ALG.c_str(), required_argument, &lraItpAlg, 0},
+            {Options::LRA_ITP_ALG.c_str(), required_argument, &lraItpAlg, 1},
```
blishko, line 84: This should not have been touched.

src/Options.cc
```diff
@@ -76,16 +81,18 @@ Options CommandLineParser::parse(int argc, char ** argv) {
...
             {Options::TPA_USE_QE.c_str(), optional_argument, &tpaUseQE, 1},
+            {Options::PROOF_FORMAT.c_str(), required_argument, nullptr, 'p'},
             {0, 0, 0, 0}
         };
+
     while (true) {
         int option_index = 0;
 
-        int c = getopt_long(argc, argv, "e:l:i:f:vh", long_options, &option_index);
+        int c = getopt_long(argc, argv, "e:l:i:f:vh:p", long_options, &option_index);
```
blishko, line 95: Shouldn't this be `p:` instead of `:p`?

src/Options.cc
```diff
@@ -161,6 +168,9 @@ Options CommandLineParser::parse(int argc, char ** argv) {
     if (tpaUseQE) {
         res.addOption(Options::TPA_USE_QE, "true");
     }
+    if (printWitness) {
+        res.addOption(Options::PRINT_WITNESS, std::to_string(printWitness));
```
blishko, line 172: Why was this moved and the argument changed?

src/Witnesses.cc
```diff
@@ -80,7 +98,7 @@ InvalidityWitness InvalidityWitness::fromErrorPath(ErrorPath const & errorPath,
     assert(vertexPredicates.size() == edgeIds.size() + 1);
     std::size_t stepCounter = 0;
     // Make the `true` the first step of the derivation
-    derivation.addDerivationStep({.index = stepCounter++, .premises = {}, .derivedFact = logic.getTerm_true(), .clauseId = {static_cast<id_t>(-1)}});
+    derivation.addDerivationStep({.index = stepCounter++, .premises = {},.derivedFact = logic.getTerm_true(), .clauseId = {static_cast<id_t>(-1)}});
```
blishko, line 101: Revert this change.

src/Witnesses.cc
```diff
@@ -111,6 +129,7 @@ void InvalidityWitness::print(std::ostream & out, Logic & logic) const {
             out << " -> ";
             for (auto index : step.premises) {
                 out << ' ' << index;
+
```
blishko, line 132: Revert this change.

src/Witnesses.h
```diff
@@ -81,6 +82,7 @@ class InvalidityWitness {
     static InvalidityWitness fromTransitionSystem(ChcDirectedGraph const & graph, std::size_t unrollings);
 
     void print(std::ostream & out, Logic & logic) const;
+
```
blishko, line 85: Revert this change.

src/Witnesses.h
```diff
@@ -143,13 +145,14 @@ class VerificationResult {
...
-    [[nodiscard]] InvalidityWitness const & getInvalidityWitness() const & { assert(answer == VerificationAnswer::UNSAFE); return std::get<InvalidityWitness>(witness); }
+    [[nodiscard]] const InvalidityWitness getInvalidityWitness() const & { assert(answer == VerificationAnswer::UNSAFE); return std::get<InvalidityWitness>(witness); }
     [[nodiscard]] std::string_view getNoWitnessReason() const & { assert(not hasWitness()); return std::get<NoWitness>(witness).getReason(); }
 
     ValidityWitness && getValidityWitness() && { assert(answer == VerificationAnswer::SAFE); return std::move(std::get<ValidityWitness>(witness)); }
     InvalidityWitness && getInvalidityWitness() && { assert(answer == VerificationAnswer::UNSAFE); return std::move(std::get<InvalidityWitness>(witness)); }
 
-    void printWitness(std::ostream & out, Logic & logic) const;
+    void printWitness(std::ostream & out, Logic & logic, ChcDirectedHyperGraph const & originalGraph, std::vector<std::shared_ptr<Term>> originalAssertions,
+                       Normalizer::Equalities const & normalizingEqualities, const std::string& format) const;
```
blishko, line 155: Style

```suggestion
                       Normalizer::Equalities const & normalizingEqualities, std::string const & format) const;
```

src/bin/golem.cpp
```diff
@@ -110,5 +109,11 @@ int main( int argc, char * argv[] ) {
             error(inputFile + " extension not recognized. File must be in smt-lib2 format (extension .smt2)");
         }
     }
+    if(options.hasOption(Options::PROOF_FORMAT)) {
```
blishko, line 112: Style
```suggestion
    if (options.hasOption(Options::PROOF_FORMAT)) {
```

src/proofs/ProofSteps.h
```diff
@@ -0,0 +1,142 @@
+//
+// Created by mambo on 8/29/23.
```
blishko, line 2: Please use the license style consistent with the rest of the project.

src/CMakeLists.txt
```diff
@@ -25,7 +25,11 @@ else(OpenSMT_FOUND)
 endif(OpenSMT_FOUND)
 
 
-add_library(golem_lib OBJECT "")
+add_library(golem_lib OBJECT ""
+        proofs/Term.h
+        proofs/Term.cpp
+        proofs/ProofSteps.h
+        proofs/ProofSteps.cpp)
```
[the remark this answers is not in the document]
M4mbo: Removing this leads to:

CMake Error at src/CMakeLists.txt:27 (target_link_libraries):
  Cannot specify link libraries for target "golem_lib" which is not built by
  this project.

src/bin/golem.cpp
```diff
@@ -110,5 +109,11 @@ int main( int argc, char * argv[] ) {
             error(inputFile + " extension not recognized. File must be in smt-lib2 format (extension .smt2)");
         }
     }
+    if (options.hasOption(Options::PROOF_FORMAT)) {
+        auto formatStr = options.getOption(Options::PROOF_FORMAT);
+        if (!(formatStr == "alethe" or formatStr == "intermediate" or formatStr == "legacy")) {
```
blishko, line 114:
```suggestion
        if (not (formatStr == "alethe" or formatStr == "intermediate" or formatStr == "legacy")) {
```

src/proofs/ProofSteps.h
```diff
@@ -0,0 +1,147 @@
+/*
+ * Copyright (c) 2023, Matias Barandiaran <matias.barandiaran03@gmail.com>
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+#ifndef GOLEM_ALETHETERMS_H
+#define GOLEM_ALETHETERMS_H
```
blishko, line 8: Guard should match the name of the file.

src/proofs/ProofSteps.h
```diff
@@ -0,0 +1,147 @@
...
+    std::vector<std::shared_ptr<Term>> originalAssertions;
+    Normalizer::Equalities const & normalizingEqualities;
+    std::ostream & out;
+    Logic & logic;
+    ChcDirectedHyperGraph originalGraph;
+
+    std::vector<Observer *> observers;
+
+    int currStep = 0;
+    int transitivityStep;
```
blishko, line 78: Should also be initialized as the other members?

src/proofs/ProofSteps.h
```diff
@@ -0,0 +1,147 @@
...
+
+    void deRegisterObserver(Observer * observer) {
+        for (int i = 0; i < observers.size(); i++) {
+            if (observer == observers[i]) { observers.erase(observers.begin() + i); }
+        }
+    }
+
+    void notifyObservers(const Step & step) {
+        for (Observer * observer : observers) { // notify all observers
+            observer->update(const_cast<Step &>(step));
```
blishko, line 142: Do not use `const_cast` here.
`update` method  of `Observer` should take `Step const &` as parameter.

src/proofs/ProofSteps.h
```diff
@@ -0,0 +1,147 @@
...
+         std::vector<std::pair<std::string, std::string>> args)
+        : type(type), stepId(stepId), clause(std::move(clause)), rule(std::move(rule)), args(std::move(args)) {}
+    Step(int stepId, stepType type, std::vector<std::shared_ptr<Term>> clause, std::string rule)
+        : type(type), stepId(stepId), clause(std::move(clause)), rule(std::move(rule)) {}
+    Step(int stepId, stepType type, std::vector<std::shared_ptr<Term>> clause)
+        : type(type), stepId(stepId), clause(std::move(clause)), rule(" ") {}
+    Step(int stepId, stepType type, std::string rule, std::vector<int> premises)
+        : type(type), stepId(stepId), rule(std::move(rule)), premises(std::move(premises)) {}
+    std::string printStepAlethe();
+    std::string printStepIntermediate();
```
blishko, line 42: These can be `const` no?
```suggestion
    std::string printStepAlethe() const;
    std::string printStepIntermediate() const;
```

src/proofs/ProofSteps.cpp
```diff
@@ -0,0 +1,909 @@
...
+ */
+
+#include "ProofSteps.h"
+#include <memory>
+#include <string>
+#include <utility>
+
+std::string Step::printStepAlethe() {
+
+    PrintVisitor printVisitor;
```
blishko, line 14:
```suggestion
```

src/proofs/ProofSteps.cpp
```diff
@@ -0,0 +1,909 @@
...
+            ss << clause[i]->printTerm();
+            if (i != clause.size() - 1) { ss << " "; }
+        }
+    }
+
+    if (type != ASSUME) { ss << ")"; }
+
+    if (rule != " ") { ss << " :rule " << rule; }
+
+    if (!premises.empty()) {
```
blishko, line 40:
```suggestion
    if (not premises.empty()) {
```

src/proofs/ProofSteps.cpp
```diff
@@ -0,0 +1,909 @@
...
+        ss << " :premises (";
+        for (int i = 0; i < premises.size(); i++) {
+            ss << "t" << premises[i];
+            if (i != premises.size() - 1) { ss << " "; }
+        }
+        ss << ")";
+    }
+
+    if (!args.empty()) {
```
blishko, line 49:
```suggestion
    if (not args.empty()) {
```

src/proofs/ProofSteps.cpp
```diff
@@ -0,0 +1,909 @@
...
+            potentialLet = originalAssertions[i - 1]->accept(&letLocatorVisitor);
+        }
+        notifyObservers(Step(currStep, Step::ASSUME,
+                             packClause(std::make_shared<Op>(
+                                 "!", packClause(originalAssertions[i - 1],
+                                                 std::make_shared<Terminal>(":named @a" + std::to_string(currStep),
+                                                                            Terminal::UNDECLARED))))));
+
+        currStep++;
+    }
```
blishko, line 435: Why do you start at 1, but then use `i-1` for access everywhere?
It seems the index can start at 0. Actually, it seems this could be turned into a range-based for loop, no?

src/proofs/Term.cpp
```diff
@@ -0,0 +1,720 @@
...
+    return printVisitor.getString();
+}
+
+void PrintVisitor::visit(Terminal * term) {
+    ss << term->getVal();
+}
+
+void PrintVisitor::visit(Op * term) {
+    ss << "(" << term->getOp();
+    for (const std::shared_ptr<Term> & arg : term->getArgs()) {
```
blishko, line 25:
```suggestion
    for (const auto & arg : term->getArgs()) {
```

src/proofs/Term.cpp
```diff
@@ -0,0 +1,720 @@
...
+    ss << ") ";
+    term->getApplication()->accept(this);
+    ss << ")";
+}
+
+std::string Op::simplifyRule() {
+    std::string op = operation;
+    if (op == "=") {
+        if ((args[0]->printTerm().find_first_not_of("( )-0123456789") == std::string::npos) and
+            (args[0]->printTerm().find_first_not_of("( )-0123456789") == std::string::npos)) {
```
blishko, line 74: Avoid calling  `printTerm` twice unnecessarily

src/proofs/Term.cpp
```diff
@@ -0,0 +1,720 @@
...
+        return "not_simplify";
+    } else if (op == "ite") {
+        return "ite_simplify";
+    } else if (op == "mod") {
+        return "mod_simplify";
+    }
+    return "Error";
+}
+
+std::string NegatedAndVisitor::visit(Op * term) {
```
blishko, line 103: It seems to me you do not need to implement this functionality in a Visitor, but a small helper function should be sufficient.

src/proofs/Term.h
```diff
@@ -0,0 +1,309 @@
...
+    virtual std::string printTerm();
+};
+
+class Terminal : public Term {
+    std::string val;
+    terminalType type;
+
+public:
+    Terminal(std::string val, terminalType t) : val(std::move(val)), type(t) {}
+    std::string getVal() { return val; }
```
blishko, line 33:
```suggestion
    std::string const & getVal() { return val; }
```

src/proofs/Term.h
```diff
@@ -0,0 +1,309 @@
...
+
+class Op : public Term {
+    std::string operation;
+    std::vector<std::shared_ptr<Term>> args;
+
+public:
+    Op(std::string opcode, std::vector<std::shared_ptr<Term>> args)
+        : operation(std::move(opcode)), args(std::move(args)) {}
+    std::string getOp() { return operation; }
+    std::vector<std::shared_ptr<Term>> getArgs() { return args; }
```
blishko, line 53:
```suggestion
    std::string const & getOp() { return operation; }
    std::vector<std::shared_ptr<Term>> const & getArgs() { return args; }
```

src/proofs/Term.h
```diff
@@ -0,0 +1,309 @@
...
+};
+
+class App : public Term {
+    std::string fun;
+    std::vector<std::shared_ptr<Term>> args;
+
+public:
+    App(std::string fun, std::vector<std::shared_ptr<Term>> args) : fun(std::move(fun)), args(std::move(args)) {}
+    std::string getFun() { return fun; }
+    std::vector<std::shared_ptr<Term>> getArgs() { return args; }
```
blishko, line 72:
```suggestion
    std::string const & getFun() { return fun; }
    std::vector<std::shared_ptr<Term>> const & getArgs() { return args; }
```

src/proofs/Term.h
```diff
@@ -0,0 +1,309 @@
...
+    std::vector<std::shared_ptr<Term>> sorts;
+    std::shared_ptr<Term> coreTerm;
+
+public:
+    Quant(std::string quant, std::vector<std::shared_ptr<Term>> vars, std::vector<std::shared_ptr<Term>> sorts,
+          std::shared_ptr<Term> coreTerm)
+        : quant(std::move(quant)), vars(std::move(vars)), sorts(std::move(sorts)), coreTerm(std::move(coreTerm)) {}
+    std::string getQuant() { return quant; }
+    std::vector<std::shared_ptr<Term>> getVars() { return vars; }
+    std::vector<std::shared_ptr<Term>> getSorts() { return sorts; }
```
blishko, line 95:
```suggestion
    std::string getQuant() const & { return quant; }
    std::vector<std::shared_ptr<Term>> const & getVars() { return vars; }
    std::vector<std::shared_ptr<Term>> const & getSorts() { return sorts; }
```

src/proofs/Term.h
```diff
@@ -0,0 +1,309 @@
...
+    std::vector<std::shared_ptr<Term>> declarations;
+    std::shared_ptr<Term> application;
+
+public:
+    Let(std::vector<std::string> termNames, std::vector<std::shared_ptr<Term>> declarations,
+        std::shared_ptr<Term> application)
+        : termNames(std::move(termNames)), declarations(std::move(declarations)), application(std::move(application)) {}
+    std::vector<std::shared_ptr<Term>> getDeclarations() { return declarations; }
+    std::shared_ptr<Term> getApplication() { return application; }
+    std::vector<std::string> getTermNames() { return termNames; }
```
blishko, line 118:
```suggestion
    std::vector<std::shared_ptr<Term>> const & getDeclarations() { return declarations; }
    std::shared_ptr<Term> getApplication() { return application; }
    std::vector<std::string> const & getTermNames() { return termNames; }
```

src/proofs/Term.h
```diff
@@ -0,0 +1,309 @@
+/*
+ * Copyright (c) 2023, Matias Barandiaran <matias.barandiaran03@gmail.com>
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+#ifndef TERM_H // Replace TERM_H with the actual name of your header file
+#define TERM_H
```
blishko, line 8:
```suggestion
#ifndef GOLEM_TERM_H
#define GOLEM_TERM_H
```

blishko (commented):
Looks good!
Left some comments to polish the code.

src/proofs/Term.cpp
M4mbo: This visitor handles non linearity, i.e:

(and true true P x Q y R z) 
-> (not Px) (not Qy) (not Rz)

blishko: That's OK, I am just saying that we do not need a visitor for that, but the functionality can be implemented in a small helper function.
Because this visitor is not really visiting the whole term, it really is only an operation on the top level of an and-term.

src/proofs/ProofSteps.h
```diff
@@ -106,8 +106,8 @@ class StepHandler {
                 Normalizer::Equalities const & normalizingEqualities, std::ostream & out, Logic & logic,
                 ChcDirectedHyperGraph originalGraph)
         : derivation(std::move(derivation)), originalAssertions(std::move(originalAssertions)),
-          normalizingEqualities(std::move(normalizingEqualities)), originalGraph(std::move(originalGraph)), out(out),
-          logic(logic) {}
+          normalizingEqualities(normalizingEqualities), originalGraph(std::move(originalGraph)), out(out),
+          logic(logic), currStep(0), renamingTransIndex(0), renamingCongIndex(0) {}
```
blishko, line 110: I actually preferred the previous version, where the members were initialized inline.
And what about the `transitivityStep`?

src/proofs/Term.cpp
```diff
@@ -71,7 +71,7 @@ std::string Op::simplifyRule() {
     std::string op = operation;
     if (op == "=") {
         if ((args[0]->printTerm().find_first_not_of("( )-0123456789") == std::string::npos) and
-            (args[0]->printTerm().find_first_not_of("( )-0123456789") == std::string::npos)) {
+            (args[1]->printTerm().find_first_not_of("( )-0123456789") == std::string::npos)) {
```
blishko, line 74: Wait, so this was a bug?
Can you come up with a test case that would expose it in the previous version?

src/proofs/ProofSteps.cpp
blishko: So, is there a reason why this cannot be a range-based for loop?

M4mbo: I am changing the contents of the array for later, that is why I need the index to modify them.

src/proofs/ProofSteps.h
M4mbo: It doesn't really need  a default initialization.

src/proofs/Term.cpp
M4mbo: Yes, it probably got changed after I modified the print function. Here I am basically checking if both arguments of the equality are numbers. Hence the index 0 and 1.

src/proofs/ProofSteps.cpp
blishko: But you are only ever using the index to access the elements in `originalAssertions`.
So you can use `for (auto & assertion : originalAssertions) { ... }`

M4mbo: In line 425 I am modifying the array's content. I understand the range-based for loop, I am just not sure if I can modify the contents of the array directly by changing the "auto & assertion" variable.

PR #46, opened 2023-09-27
Proof length reduction

m4mbo:
Changed simplification algorithm from linear cong propagation to a recursive depth-first post order procedure (operation arguments will be simplified before propagating the chain).

### DIFF
--- a/src/proofs/ProofSteps.cpp
+++ b/src/proofs/ProofSteps.cpp
@@ -256,12 +256,13 @@ void StepHandler::buildAletheProof() {
         directSimplification(requiredMP, implicationStep, lastClause, renamedImpLHS);
     }
 
-    notifyObservers(
-        Step(currentStep, Step::STEP, packClause(std::make_shared<Terminal>("(not false)", Term::UNDECLARED)), "false"));
+    notifyObservers(Step(currentStep, Step::STEP,
+                         packClause(std::make_shared<Terminal>("(not false)", Term::UNDECLARED)), "false"));
 
     currentStep++;
     // Get empty clause
-    notifyObservers(Step(currentStep, Step::STEP, "resolution", std::vector<std::size_t>{currentStep - 2, currentStep - 1}));
+    notifyObservers(
+        Step(currentStep, Step::STEP, "resolution", std::vector<std::size_t>{currentStep - 2, currentStep - 1}));
 }
 
 void StepHandler::instantiationSteps(std::size_t i) {
@@ -296,8 +297,8 @@ void StepHandler::instantiationSteps(std::size_t i) {
 
         currentStep++;
 
-        notifyObservers(
-            Step(currentStep, Step::STEP, packClause(unusedRem), "resolution", std::vector<std::size_t>{quantStep, currentStep - 1}));
+        notifyObservers(Step(currentStep, Step::STEP, packClause(unusedRem), "resolution",
+                             std::vector<std::size_t>{quantStep, currentStep - 1}));
 
         currentStep++;
 
@@ -411,8 +412,9 @@ void StepHandler::directSimplification(std::vector<std::size_t> requiredMP, std:
     }
 }
 
-void StepHandler::conjunctionSimplification(std::vector<std::size_t> requiredMP, std::shared_ptr<Term> const & lastClause,
-                                            std::size_t implicationStep, std::shared_ptr<Term> const & renamedImpLHS) {
+void StepHandler::conjunctionSimplification(std::vector<std::size_t> requiredMP,
+                                            std::shared_ptr<Term> const & lastClause, std::size_t implicationStep,
+                                            std::shared_ptr<Term> const & renamedImpLHS) {
 
     auto termToSimplify = std::dynamic_pointer_cast<Op>(lastClause)->getArgs()[0];
     auto simplification = std::dynamic_pointer_cast<Op>(lastClause)->getArgs()[1];
--- a/src/proofs/ProofSteps.cpp
+++ b/src/proofs/ProofSteps.cpp
@@ -169,6 +169,32 @@ void StepHandler::buildIntermediateProof() {
 
 void StepHandler::buildAletheProof() {
 
+    // Getting the instantiated variable-value pairs
+    std::vector<std::pair<std::string, std::string>> instPairs =
+        getInstPairs(1, normalizingEqualities[0]);
+    InstantiateVisitor instantiateVisitor(instPairs);
+
+    CongChainVisitor congChainVisitor(currStep);
+
+    currTerm = std::dynamic_pointer_cast<Quant>(originalAssertions[0])->getCoreTerm();
+
+    std::cout << currTerm->printTerm() << "\n\n";
+
+    currTerm = currTerm->accept(&instantiateVisitor);
+
+    std::cout << currTerm->printTerm() << "\n\n";
+
+    currTerm->accept(&congChainVisitor);
+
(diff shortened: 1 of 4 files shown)


src/proofs/ProofSteps.cpp
```diff
@@ -178,17 +178,14 @@ void StepHandler::buildAletheProof() {
...
         if (step.premises.empty()) { continue; }
 
         std::vector<int> requiredMP;
 
         if (!modusPonensSteps.empty()) {
             // Get the necessary steps for modus ponens
-            for (std::size_t i = 0; i < step.premises.size(); i++) {
-                requiredMP.push_back(modusPonensSteps[(int)step.premises[i] - 1]);
+            for (unsigned long premise : step.premises) {
+                requiredMP.push_back(modusPonensSteps[(int)premise - 1]);
```
blishko, line 188: Do not use C-style casts.

Looking at this code, I think in several places, we should be using `std::size_t` instead of `int`.
For example, `modusPonensSteps` should be `std::vector<std::size_t>>` not vector of `ints`.
Similarly, `currStep` should be `std::size_t`, it can never be negative.
While at it, please rename `currStep` to `currentStep`.

blishko: Also, just use `auto premise` here, instead of `unsigned long premise`.

src/proofs/ProofSteps.cpp
```diff
@@ -199,7 +196,7 @@ void StepHandler::buildAletheProof() {
 
         std::shared_ptr<Op> implication = std::dynamic_pointer_cast<Op>(currTerm);
 
-        implicationLHS = implication->getArgs()[0];
+        implicationLHS = implication->getArgs()[0]->accept(&copyVisitor);
```
blishko, line 199: What is the reason for a copy here?

src/proofs/ProofSteps.cpp
```diff
@@ -869,7 +583,7 @@ StepHandler::getInstPairs(int stepIndex, vec<Normalizer::Equality> const & stepN
     return res;
 }
 
-int StepHandler::stepReusage(std::shared_ptr<Term> term) {
+int StepHandler::stepReusage(const std::shared_ptr<Term> & term) {
```
blishko, line 586: Why is `stepReusage` returning an `int`, but the callers of this method never do anything with this return value?
Can it be changed to `void`?

src/proofs/ProofSteps.h
```diff
@@ -74,26 +74,16 @@ class StepHandler {
...
     std::vector<int> stepsToReuse = {-1, -1, -1};
 
     std::shared_ptr<Term> implicationRHS;
     std::shared_ptr<Term> implicationLHS;
     std::shared_ptr<Term> currTerm;
-    std::vector<int> modusPonensSteps;    // Modus Ponens Steps to derive the next node
-    std::vector<int> simplificationSteps; // Simplification steps for final resolution of LHS
-
-    std::shared_ptr<Term> originalLhsPrimaryBranch;
+    std::vector<int> modusPonensSteps; // Modus Ponens Steps to derive the next node
```
blishko, line 83:
```suggestion
    std::vector<std::size_t> modusPonensSteps; // Modus Ponens Steps to derive the next node
```

src/proofs/Term.cpp
```diff
@@ -192,12 +292,12 @@ std::shared_ptr<Term> RemoveUnusedVisitor::visit(Quant * term) {
...
         auto var = vars[i];
-        for (auto varInUse : varsInUse) {
+        for (const auto & varInUse : varsInUse) {
             if (var->printTerm() == varInUse) { inUse = true; }
         }
         if (!inUse) {
-            vars.erase(vars.begin() + i);
-            sorts.erase(sorts.begin() + i);
+            vars.erase(vars.begin() + int(i));
+            sorts.erase(sorts.begin() + int(i));
```
blishko, line 300: I don't see the need to cast this to `int`. What am I missing?
```suggestion
            vars.erase(vars.begin() + i);
            sorts.erase(sorts.begin() + i);
```

src/proofs/Term.h
```diff
@@ -16,10 +16,8 @@ class Term {
     enum terminalType { VAR, REAL, INT, SORT, BOOL, UNDECLARED };
     virtual termType getTermType() = 0;
     virtual terminalType getTerminalType() = 0;
-    virtual void accept(class PrintVisitor *) = 0;
+    virtual void accept(class VoidVisitor *) = 0;
```
blishko, line 19: I don't think `class` is needed here.
```suggestion
    virtual void accept(VoidVisitor *) = 0;
```

src/proofs/Term.h
```diff
@@ -57,12 +53,12 @@ class Op : public Term {
     std::string simplifyRule();
     bool nonLinearity();
     std::string nonLinearSimplification();
+    void setArg(int i, std::shared_ptr<Term> newArg) { args[i] = std::move(newArg); }
```
blishko, line 56: This method is no longer needed, no?

src/proofs/Term.h
```diff
@@ -210,46 +182,53 @@ class SimplifyVisitor : public LogicVisitor {
...
-    virtual bool visit(Let *) = 0;
-};
-
-class IsPrimaryBranchVisitor : public BooleanVisitor {
-    Term * branch;
+class CongChainVisitor : public LogicVisitor {
+    int transCase = 0; // 0 for regular case, 1 for trans after ">="
+    InstantiateVisitor copyVisitor;
+    int currStep;
+    class simpleStep {
```
blishko, line 212: Name of the type should start with capital letter.

blishko (commented):
Looks good! Just some small cleanup would be good.

One comment related to style is that we use east-const style.
Please change expression `const <type> &` to `<type> const &`.
(See this post on this topic: https://mariusbancila.ro/blog/2018/11/23/join-the-east-const-revolution/)

src/proofs/Term.cpp
m4mbo: It gives a warning about narrowing conversion.

src/proofs/Term.h
m4mbo: You need to keep it as the class is only defined later.

src/proofs/Term.h
m4mbo: We are still using it, but on a copy, to avoid simplifying the actual term.

src/proofs/Term.cpp
blishko: OK, I see. While looking at this code, I noticed that it actually does not work correctly.
I'll write a suggestion in a separate comment.

src/proofs/Term.cpp
```diff
@@ -192,12 +293,12 @@ std::shared_ptr<Term> RemoveUnusedVisitor::visit(Quant * term) {
...
         }
         inUse = false;
     }
```
blishko, line 304: This loop does not really work correctly, because if there are two consecutive unused variables, this loop will skip the second one.

Here is an example that generated invalid proof according to carcara:
```
(set-logic HORN)
(declare-fun P (Int) Bool)

(assert (forall ((x Int) (unused Int) (unused2 Bool)) (=> (= x 0) (P x))))

(assert (forall ((x Int)) (=> (P x) false)))


(check-sat)
(exit)
```

If `inUse` is `false`, `i` should not be incremented, because `erase` moves the rest of the elements one place lower.

blishko (approved):
OK, looks good!
Thank you for this work!

PR #47, opened 2023-10-07
DAG-fix: unique restrictions for every edge

BritikovKI:


blishko:
It seems this can all be squashed in a single commit, no?
Can you please add the description to the PR? Explain what is the problem and what is your proposed solution.

I am trying to review the changes, but I am not sure I get the details. What is the relation between `blocked_children` and `checked_children`?

PR #48, opened 2023-10-08
DAG-disj: Fixed conjunction to disjunction in "addRestrictions"

BritikovKI:
### Description:
The conjunction of constraints(which was used previously), restricted the paths which should not've been restricted, leading to the spurious SAFE results.

### Solution:
- Conjunction was replaced with disjunction, therefore as long as at least one branch is reachable from the end of TS, we still should consider the node as a potential candidate.
- Two tests are added to validate this specific property.


blishko:
I guess it looks good :)

Can you just specify in the PR description what was the problem and what the fix means?

PR #50, opened 2023-11-13
Options: added an option to turn off the TS DAG solving

BritikovKI:
Added an option that allows to turn off the DAG of TS algorithm for the solving.


blishko (approved):
LGTM!

PR #60, opened 2024-02-12
DAG Safety witness

BritikovKI:
This pull request adds the solver capacity to produce validity witness for the DAGs of Transition Systems solving algorithms.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,7 @@
 build
 
-cmake-build-release
-
 ## idea
-cmake-build-debug
+cmake-build-*
 .idea
 *.log
 run_benchmark.sh
--- a/src/engine/TPA.cc
+++ b/src/engine/TPA.cc
@@ -12,6 +12,7 @@
 #include "TermUtils.h"
 #include "TransformationUtils.h"
 #include "TransitionSystem.h"
+#include "Witnesses.h"
 #include "transformers/BasicTransformationPipelines.h"
 #include "transformers/SingleLoopTransformation.h"
 #include "utils/SmtSolver.h"
@@ -23,6 +24,10 @@
 
 const std::string TPAEngine::TPA = "tpa";
 const std::string TPAEngine::SPLIT_TPA = "split-tpa";
+
+bool TPAEngine::shouldComputeWitness() const {
+    return options.hasOption(Options::COMPUTE_WITNESS);
+}
 
 std::unique_ptr<TPABase> TPAEngine::mkSolver() {
     assert(options.hasOption(Options::ENGINE));
@@ -1515,6 +1520,8 @@ private:
 
     InvalidityWitness computeInvalidityWitness() const;
 
+    std::variant<NoWitness, ValidityWitness> computeValidityWitness();
+
     void addRestrictions(SymRef node, PTRef fla);
 
     void updateRestrictions(SymRef node);
@@ -1589,8 +1596,17 @@ VerificationResult TransitionSystemNetworkManager::solve() && {
     while (true) {
         if (getNode(current).blocked_children == getNode(current).children.size()) {
             if (current == graph.getEntry()) {
-                // TODO: Compute witness
-                return {VerificationAnswer::SAFE, NoWitness{"Unable to compute witness from TPA for DAG of TSs yet"}};
+                if (this->owner.shouldComputeWitness()) {
+                    auto witness = computeValidityWitness();
+                    if (std::holds_alternative<ValidityWitness>(witness)) {
+                        return {VerificationAnswer::SAFE, std::get<ValidityWitness>(witness)};
+                    } else {
+                        return {VerificationAnswer::SAFE,
+                                NoWitness("Unexpected situation occurred during witness computation in TPA engine")};
+                    }
+                } else {
+                    return {VerificationAnswer::SAFE, NoWitness()};
+                }
             }
             getNode(current).trulyReached = PTRef_Undef;
             getNode(current).blocked_children = 0;
@@ -1689,6 +1705,40 @@ InvalidityWitness TransitionSystemNetworkManager::computeInvalidityWitness() con
     InvalidityWitness witness;
     ErrorPath path(std::move(errorPath));
     return InvalidityWitness::fromErrorPath(path, graph);
+}
+
+std::variant<NoWitness, ValidityWitness> TransitionSystemNetworkManager::computeValidityWitness() {
+    assert(isTransitionSystemDAG(graph));
+    auto vertices = graph.getVertices();
+    TermUtils utils(logic);
+    TimeMachine timeMachine(logic);
+    TermUtils::substitutions_map subs;
+    ValidityWitness::definitions_t definitions;
+
+    for (auto vertex : vertices) {
+        if (vertex == graph.getEntry() || vertex == graph.getExit()) continue;
+
+        auto graphVars = utils.predicateArgsInOrder(graph.getStateVersion(vertex));
+        vec<PTRef> unversionedVars;
+        auto systemVars = networkMap.at(vertex).solver->getStateVars(0);
+
+        for (std::size_t i = 0; i < graphVars.size(); ++i) {
+            unversionedVars.push(timeMachine.getUnversioned(graphVars[i]));
+            subs.insert({systemVars[i], unversionedVars.last()});
+        }
+        getNode(vertex).solver->resetInitialStates(getNode(vertex).trulySafe);
+        // It uses the query which was set during solving of DAG
+        auto [res, explanation] = queryTransitionSystem(networkMap.at(vertex));
+        assert(res == ReachabilityResult::UNREACHABLE);
+        if (res == ReachabilityResult::UNREACHABLE) {
+            PTRef graphInvariant = utils.varSubstitute(getNode(vertex).solver->getInductiveInvariant(), subs);
+            PTRef unversionedPredicate = logic.mkUninterpFun(vertex, std::move(unversionedVars));
+            definitions[unversionedPredicate] = graphInvariant;
+        } else {
+            return NoWitness();
+        }
+    }
+    return ValidityWitness(std::move(definitions));
 }
 
 TransitionSystem TransitionSystemNetworkManager::constructTransitionSystemFor(SymRef vid) const {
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -1,9 +1,9 @@
 build
-
 cmake-build-release
 
 ## idea
 cmake-build-debug
+cmake-build-debug-event-trace
```
blishko, line 6: I think we use `cmake-build-*` to replace all three entries that fall under regex.

src/engine/TPA.cc
```diff
@@ -1515,6 +1516,8 @@ class TransitionSystemNetworkManager {
 
     InvalidityWitness computeInvalidityWitness() const;
 
+    ValidityWitness computeValidityWitness();
```
blishko, line 1519: Can it be `const`?

src/engine/TPA.cc
```diff
@@ -1590,7 +1593,7 @@ VerificationResult TransitionSystemNetworkManager::solve() && {
         if (getNode(current).blocked_children == getNode(current).children.size()) {
             if (current == graph.getEntry()) {
                 // TODO: Compute witness
-                return {VerificationAnswer::SAFE, NoWitness{"Unable to compute witness from TPA for DAG of TSs yet"}};
+                return {VerificationAnswer::SAFE, computeValidityWitness()};
```
blishko, line 1596: You can now remove the TODO.

src/engine/TPA.cc
```diff
@@ -1691,6 +1694,40 @@ InvalidityWitness TransitionSystemNetworkManager::computeInvalidityWitness() con
     return InvalidityWitness::fromErrorPath(path, graph);
 }
 
+ValidityWitness TransitionSystemNetworkManager::computeValidityWitness() {
+    if (not isTransitionSystemDAG(graph)) { return {}; }
```
blishko, line 1698: This should always be true, no? We can `assert` it if needed.

src/engine/TPA.cc
```diff
@@ -1691,6 +1694,40 @@ InvalidityWitness TransitionSystemNetworkManager::computeInvalidityWitness() con
...
+    auto vertices = graph.getVertices();
+    assert(vertices.size() >= 3);
```
blishko, line 1700: What is the reason for this `assert`?

src/engine/TPA.cc
```diff
@@ -1691,6 +1694,40 @@ InvalidityWitness TransitionSystemNetworkManager::computeInvalidityWitness() con
...
+    TermUtils utils(logic);
+    TimeMachine timeMachine(logic);
+    TermUtils::substitutions_map subs;
+    std::vector<ValidityWitness> witnesses;
```
blishko, line 1704: This is not used an can be removed.

src/engine/TPA.cc
```diff
@@ -1691,6 +1694,40 @@ InvalidityWitness TransitionSystemNetworkManager::computeInvalidityWitness() con
...
+    ValidityWitness::definitions_t definitions;
+
+    for (auto node : vertices) {
+        auto vertex = node;
```
blishko, line 1708:
```suggestion
    for (auto vertex : vertices) {
```

src/engine/TPA.cc
```diff
@@ -1691,6 +1694,40 @@ InvalidityWitness TransitionSystemNetworkManager::computeInvalidityWitness() con
...
+        auto graphVars = utils.predicateArgsInOrder(graph.getStateVersion(vertex));
+        vec<PTRef> unversionedVars;
+        auto systemVars = networkMap.at(vertex).solver->getStateVars(0);
+
+        for (std::size_t i = 0; i < graphVars.size(); ++i) {
+            unversionedVars.push(timeMachine.getUnversioned(graphVars[i]));
+            subs.insert({systemVars[i], unversionedVars.last()});
+        }
+        getNode(vertex).solver->resetInitialStates(getNode(vertex).trulySafe);
+        auto [res, explanation] = queryTransitionSystem(networkMap.at(vertex));
```
blishko, line 1720: Maybe add a note here that this uses the query that was set in the main solving algorithm?

src/engine/TPA.cc
```diff
@@ -1691,6 +1694,40 @@ InvalidityWitness TransitionSystemNetworkManager::computeInvalidityWitness() con
...
+        assert(res == ReachabilityResult::UNREACHABLE);
+        if (res == ReachabilityResult::UNREACHABLE) {
```
blishko, line 1722: I think we should add an `else` case where we return `NoWitness` and explain what went wrong.

blishko (commented):
Looks good!

Can you add an explanation why this guarantees that the computed invariants will satisfy the edge constraints (in the sense that source invariant together with edge constraint implies the target invariant)?

blishko:
Also, I think we can squash all to a single commit, no?

src/engine/TPA.cc
BritikovKI: Unfortunately no, because `queryTransitionSystem` is called inside this function

src/engine/TPA.cc
BritikovKI: To be honest, is should always be true, if it is REACHABLE, it means there is a problem in the solving algorithm(as it wasn't able to cover all passes)

The reason I've even added if, is that CI throws an error that `res` is not used otherwise

src/engine/TPA.cc
BritikovKI: Done

src/engine/TPA.cc
BritikovKI: Changed to assert

src/engine/TPA.cc
BritikovKI: Removed

src/engine/TPA.cc
BritikovKI: Done

src/engine/TPA.cc
BritikovKI: Done

src/engine/TPA.cc
BritikovKI: Done

src/engine/TPA.cc
blishko: I understand that it SHOULD be true :)
I agree that if it is not, something has gone wrong, so we should not pretend that we have the witness, and instead be safe and return no witness.
It can be something like `return NoWitness("Unexpected situation occurred during witness computation in TPA engine");`

.gitignore
```diff
@@ -1,9 +1,8 @@
 build
-
 cmake-build-release
```
blishko, line 2:
```suggestion
```
This can go away now, it is covered by the `*` version.

src/engine/TPA.cc
```diff
@@ -1588,10 +1591,7 @@ VerificationResult TransitionSystemNetworkManager::solve() && {
     activePath.clear();
     while (true) {
         if (getNode(current).blocked_children == getNode(current).children.size()) {
-            if (current == graph.getEntry()) {
-                // TODO: Compute witness
-                return {VerificationAnswer::SAFE, NoWitness{"Unable to compute witness from TPA for DAG of TSs yet"}};
-            }
+            if (current == graph.getEntry()) { return {VerificationAnswer::SAFE, computeValidityWitness()}; }
```
blishko, line 1594: We actually need to check if we need to compute witnesses or not.
If not, there is not point in trying to compute it, it would actually be harmful.
That concerns both valid and invalid case.

BritikovKI: Oh, you mean specifically if there is a demand to compute witness or not...

BritikovKI: I don't see much harm, but slowdown is indeed guaranteed(basically if no validate option, we can return noWitness everywhere)

I would move it to the different PR though, as at is a different issue, connected not only with our algorithm/TPA, but I suppose also with other engines

.gitignore
BritikovKI: Updated

src/engine/TPA.cc
blishko: By harmful I meant that some instances might not be solved anymore because of the witness computation, even though the witness was not required.

It should be fixed here for exactly that reason. The change is simple, you just need to check if witness computation was requested and only call `computeValidityWitness` in such case.
TPA for a transition system already behaves that way.

src/engine/TPA.cc
BritikovKI: I agree with general idea, I just don't like the idea of `computeValidityWitness` returning `variant<NoWitness, ValidityWitness>` 😅

blishko: I see, OK. How about immediately returning empty `ValidityWitness`?

src/engine/TPA.cc
BritikovKI: The thing is, this issue is uniform, it also exists for Spacer, Lawi, etc, because we always produce witnesses for SAFE/UNSAFE. Therefore imho it should be a uniform change all over the project, that is somewhat independent from this specific DAG approach. I've added a check, but I think it should be general tbh...

src/engine/TPA.cc
```diff
@@ -1589,8 +1592,16 @@ VerificationResult TransitionSystemNetworkManager::solve() && {
     while (true) {
         if (getNode(current).blocked_children == getNode(current).children.size()) {
             if (current == graph.getEntry()) {
-                // TODO: Compute witness
-                return {VerificationAnswer::SAFE, NoWitness{"Unable to compute witness from TPA for DAG of TSs yet"}};
+                if (Options::COMPUTE_WITNESS == "true") {
```
blishko, line 1595: This is not going to work, `Options::COMPUTE_WITNESS` is just the string `"compute-witness"`, so this comparison is always `false`.  `TPAEngine` has `options` field, which you can query.
You will have to expose it somehow. I suggest to add a helper public method `bool shouldComputeWitness() const`. It can also be used internally in other places in TPA as well.

src/engine/TPA.cc
```diff
@@ -1589,8 +1592,16 @@ VerificationResult TransitionSystemNetworkManager::solve() && {
...
+                    auto witness = computeValidityWitness();
+                    if (holds_alternative<ValidityWitness>(witness)) {
+                        return {VerificationAnswer::SAFE, get<ValidityWitness>(witness)};
+                    } else {
+                        return {VerificationAnswer::SAFE, NoWitness()};
```
blishko, line 1600:
```suggestion
                        return {VerificationAnswer::SAFE, std::get<NoWitness>(witness)};
```

Otherwise you would be throwing away the reason why no witness was computed.

src/engine/TPA.cc
BritikovKI: Ok, thanks!

blishko (approved):
LGTM!

PR #61, opened 2024-03-24
SingleLoopTransform: Deal with alien variables in validity witnesses in backtranslation

### LINKED ISSUE #45: Figure out how to properly do the invariant backtranslation in transformation to single TS
For example `chc22/LIA-Lin/chc-LIA-Lin_082.smt2` for engine `imc` currently runs into alien variables in vertex invariant and is therefore unable to proceed with the backtranslation.

blishko:
In the current encoding to single transition system, when translating edge constraint, we do not put any restriction on the big system's state variables that do no correspond to the variables of the target node. This means that when we encounter alien variables during backtranslation of validity witness, we can universally quantify these variables to obtain predicate representation only over its arguments. We can then apply quantifier elimination.
Since QE works on existential quantifiers, we have to first negate the original invariant and then negate again after QE.

Fixes #45.

### DIFF
--- a/src/transformers/SingleLoopTransformation.cc
+++ b/src/transformers/SingleLoopTransformation.cc
@@ -6,6 +6,7 @@
 
 #include "SingleLoopTransformation.h"
 
+#include "QuantifierElimination.h"
 #include "TransformationUtils.h"
 #include "utils/SmtSolver.h"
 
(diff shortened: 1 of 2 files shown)


BritikovKI (approved):
Lgtmt

PR #80, opened 2024-08-29
TPA: Merge DAG nested loops

BritikovKI:
This is an implementation of the merging of the nested loops into a single Transition System. It is done via the same approach as the transformation of the whole instance into the single System, but is adapted to only merge loops in the Graph structure into a single node.

### DIFF
--- a/src/TransformationUtils.cc
+++ b/src/TransformationUtils.cc
@@ -134,6 +134,7 @@ std::vector<EId> detectLoop(const ChcDirectedGraph & graph) {
     std::unordered_set<SymRef, SymRefHash> visitedVertices;
     std::unordered_set<SymRef, SymRefHash> verticesOnStack;
     visit(visitedVertices, verticesOnStack, graphRepresentation, graph, graph.getEntry(), loop);
+    std::reverse(loop.begin(), loop.end());
     return loop;
 }
 
--- a/src/engine/TPA.cc
+++ b/src/engine/TPA.cc
@@ -14,6 +14,7 @@
 #include "TransitionSystem.h"
 #include "Witnesses.h"
 #include "transformers/BasicTransformationPipelines.h"
+#include "transformers/NestedLoopTransformation.h"
 #include "transformers/SingleLoopTransformation.h"
 #include "utils/SmtSolver.h"
 
@@ -73,8 +74,17 @@ VerificationResult TPAEngine::solve(const ChcDirectedGraph & graph) {
                 assert(false);
                 throw std::logic_error("Unreachable!");
         }
-    } else if (isTransitionSystemDAG(graph) && not options.hasOption(Options::FORCE_TS)) {
-        return solveTransitionSystemGraph(graph);
+    } else if ((isTransitionSystemDAG(graph) && not options.hasOption(Options::FORCE_TS)) ||
+               options.hasOption(Options::SIMPLIFY_NESTED)) {
+        if (options.hasOption(Options::SIMPLIFY_NESTED)) {
+            NestedLoopTransformation transformation;
+            auto [transformedGraph, preTranslator] = transformation.transform(graph);
+            assert(isTransitionSystemDAG(*transformedGraph));
+            auto res = solveTransitionSystemGraph(*transformedGraph);
+            return preTranslator->translate(res);
+        } else {
+            return solveTransitionSystemGraph(graph);
+        }
     }
     // Translate CHCGraph into transition system
     SingleLoopTransformation transformation;
--- a/src/graph/ChcGraph.cc
+++ b/src/graph/ChcGraph.cc
@@ -6,6 +6,8 @@
 
 #include "ChcGraph.h"
 
+#include "transformers/CommonUtils.h"
+#include "TransformationUtils.h"
 #include <iostream>
 #include <map>
 
@@ -599,4 +601,126 @@ std::unique_ptr<ChcDirectedHyperGraph> ChcDirectedHyperGraph::makeEmpty(Logic & 
         std::move(predicateRepresentation),
         logic
         );
+}
+
+WitnessInfo ChcDirectedGraph::contractConnectedVertices(std::vector<EId> edges) {
+    assert(edges.size() >= 2);
+    assert(getSource(edges[0]) == getTarget(edges[edges.size() - 1]));
+    TimeMachine timeMachine(logic);
+    TermUtils utils(logic);
+    std::vector<SymRef> vertices;
+    assert(std::adjacent_find(edges.begin(), edges.end(), [&](auto first, auto second) { return getTarget(first) != getSource(second); }) == edges.end());
+    for (uint i = 0; i < edges.size(); i++) {
+        vertices.push_back(getSource(edges[i]));
+    }
+
+    LocationVarMap locationVars;
+    PositionVarMap argVars;
+    locationVars.reserve(vertices.size());
+    for (auto vertex : vertices) {
+        auto varName = std::string(".loc_") + std::to_string(vertex.x);
+        locationVars.insert({vertex, timeMachine.getVarVersionZero(varName, logic.getSort_bool())});
+    }
+
+    for (auto vertex : vertices) {
+        auto args_count = logic.getSym(vertex).nargs();
+        for (uint32_t i = 0; i < args_count; ++i) {
+            VarPosition pos = {vertex, i};
+            auto varName = std::string(".arg_") + std::to_string(vertex.x) + '_' + std::to_string(i);
+            PTRef var = timeMachine.getVarVersionZero(varName, logic.getSym(vertex)[i]);
+            argVars.insert({pos, var});
+        }
+    }
+
+    vec<PTRef> transitionRelationComponent;
+
+    // KB: creation of the location variables for nodes and argument variables (params of predicates)
+    std::vector<EId> allEdges = getEdges();
+    EdgeTranslator edgeTranslator{*this, locationVars, argVars, {}};
+    // KB: Translation of self-loop edges for the edges vertices
+    //     and edges between vertices in DAG
+    for (auto edge : allEdges) {
+        if (std::find(vertices.begin(), vertices.end(), getSource(edge)) != vertices.end() &&
+            std::find(vertices.begin(), vertices.end(), getTarget(edge)) != vertices.end()) {
+            transitionRelationComponent.push(edgeTranslator.translateEdge(getEdge(edge)));
+            removeEdge(edge);
+        }
+    };
+    PTRef transitionRelation = logic.mkOr(std::move(transitionRelationComponent));
+
+    // KB: Extraction of the state variables of the nested loop nodes, creation of new predicate to replace node with
+    //     nested loops
+    std::vector<PTRef> stateVars = [&locationVars, &argVars]() {
+        std::vector<PTRef> ret;
+        for (auto && entry : locationVars) {
+            ret.push_back(entry.second);
+        }
+        for (auto && entry : argVars) {
+            ret.push_back(entry.second);
+        }
+        return ret;
+    }();
+
+    vec<SRef> argSorts;
+    std::vector<PTRef> vars;
+    for (auto arg : stateVars) {
+        argSorts.push(logic.getSortRef(arg));
+        vars.push_back(timeMachine.getUnversioned(arg));
+    }
+
+    std::string name;
+    for (auto vertex : vertices) {
+        name += logic.getSymName(vertex);
+    }
+
+    SymRef newRef = logic.declareFun(name, logic.getSort_bool(), argSorts);
+    predicates.addRepresentation(newRef, std::move(vars));
+
+    // KB: Changing input and output edges to and from the loop in the graph with new state and loc variables
+    //     Redirecting edges to the newly created vertice
+    allEdges = getEdges();
+    for (auto edge : allEdges) {
+        if (std::find(vertices.begin(), vertices.end(), getSource(edge)) != vertices.end()) {
+            PTRef newSourceLoc = locationVars.at(getSource(edge));
+            auto oldEdgeVars = getVariablesFromEdge(logic, *this, edge);
+            std::unordered_map<PTRef, PTRef, PTRefHash> subMap;
+            for (uint i = 0; i < oldEdgeVars.stateVars.size(); i++) {
+                subMap.insert(std::make_pair(oldEdgeVars.stateVars[i], argVars[{getSource(edge), i}]));
+            }
+            updateEdgeSource(edge, newRef);
+            PTRef label = utils.varSubstitute(logic.mkAnd(getEdgeLabel(edge), newSourceLoc), subMap);
+
+            updateEdgeLabel(edge, InterpretedFla{label});
+        }
+        if (std::find(vertices.begin(), vertices.end(), getTarget(edge)) != vertices.end()) {
+
+            PTRef newTargetLoc = [&]() -> PTRef {
+                vec<PTRef> negatedLocations;
+                negatedLocations.capacity(locationVars.size());
+                for (auto && entry : locationVars) {
+                    if (entry.first != getTarget(edge)) {
+                        negatedLocations.push(logic.mkNot(timeMachine.sendVarThroughTime(entry.second, 1)));
+                    }
+                }
+                return logic.mkAnd(std::move(negatedLocations));
+            }();
+            PTRef destination = timeMachine.sendVarThroughTime(locationVars.at(getTarget(edge)), 1);
+
+            auto oldEdgeVars = getVariablesFromEdge(logic, *this, edge);
+            std::unordered_map<PTRef, PTRef, PTRefHash> subMap;
+            for (uint i = 0; i < oldEdgeVars.nextStateVars.size(); i++) {
+                subMap.insert(std::make_pair(oldEdgeVars.nextStateVars[i],
+                                             timeMachine.sendVarThroughTime(argVars[{getTarget(edge), i}], 1)));
+            }
+            updateEdgeTarget(edge, newRef);
+            PTRef label = utils.varSubstitute(getEdgeLabel(edge), subMap);
+
+            updateEdgeLabel(edge, InterpretedFla{rewriteMaxArityClassic(
+                                            logic, logic.mkAnd({newTargetLoc, destination, label}))});
+        }
+    }
+
+    // KB: Creating new self-looping edge (which represents all of the transitions in the outer loop)
+    newEdge(newRef, newRef, InterpretedFla{transitionRelation});
+    return {newRef, locationVars, argVars};
 }
--- a/src/graph/ChcGraph.h
+++ b/src/graph/ChcGraph.h
@@ -107,6 +107,28 @@ struct VertexHasher {
     }
 };
 
+struct VarPosition {
+    SymRef vertex;
+    uint32_t pos;
+
+    inline bool operator==(VarPosition other) const { return vertex == other.vertex and pos == other.pos; }
+};
+struct VarPositionHasher {
+    std::size_t operator()(VarPosition pos) const {
+        std::hash<std::uint32_t> hasher;
+        return hasher(pos.vertex.x) ^ hasher(pos.pos);
+    }
+};
+
+using LocationVarMap = std::unordered_map<SymRef, PTRef, SymRefHash>;
+using PositionVarMap = std::unordered_map<VarPosition, PTRef, VarPositionHasher>;
+
+struct WitnessInfo {
+    SymRef loopVertex;
+    LocationVarMap locations;
+    PositionVarMap positions;
+};
+
 class ChcDirectedGraph {
     std::map<EId, DirectedEdge> edges;
     LinearCanonicalPredicateRepresentation predicates;
@@ -114,8 +136,10 @@ class ChcDirectedGraph {
     mutable std::size_t freeId {0};
 
     // graph transformations
-    friend class GraphTransformations;
+    friend class NestedLoopTransformation;
+    friend class SingleLoopTransformation;
     void contractVertex(SymRef sym);
+    WitnessInfo contractConnectedVertices(std::vector<EId> edges);
     void mergeEdges(EId incoming, EId outgoing);
     void deleteNode(SymRef sym);
     PTRef mergeLabels(DirectedEdge const & incoming, DirectedEdge const & outgoing);
@@ -158,9 +182,9 @@ public:
         return getEdge(eid).from;
     }
 
-	SymRef getTarget(EId eid) const {
-		return getEdge(eid).to;
-	}
+    SymRef getTarget(EId eid) const {
+            return getEdge(eid).to;
+    }
 
     std::unique_ptr<ChcDirectedHyperGraph> toHyperGraph() const;
 
@@ -192,6 +216,24 @@ private:
     void newEdge(SymRef from, SymRef to, InterpretedFla label) {
         EId eid = freshId();
         edges.emplace(eid, DirectedEdge{.from = from, .to = to, .fla = label, .id = eid});
+    }
+
+    void updateEdgeLabel(EId edge, InterpretedFla label) {
+        edges[edge].fla  = label;
+    }
+
+    void updateEdgeSource(EId edge, SymRef source) {
+        edges[edge].from  = source;
+    }
+
+    void updateEdgeTarget(EId edge, SymRef target) {
+        edges[edge].to  = target;
+    }
+
+    void removeEdge(EId edge){
+        const auto it = edges.find(edge);
+        if (it != edges.end())
+            edges.erase(it);
     }
 
 };
--- a/src/transformers/NestedLoopTransformation.cc
+++ b/src/transformers/NestedLoopTransformation.cc
@@ -1,0 +1,212 @@
+/*
+ * Copyright (c) 2024, Konstantin Britikov <britikovki@gmail.com>
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+#include "NestedLoopTransformation.h"
+
+#include "QuantifierElimination.h"
+#include "TransformationUtils.h"
+#include "graph/ChcGraph.h"
+#include "utils/SmtSolver.h"
+
+std::tuple<std::unique_ptr<ChcDirectedGraph>, std::unique_ptr<NestedLoopTransformation::WitnessBackTranslator>>
+NestedLoopTransformation::transform(ChcDirectedGraph const & graph) {
+    auto vertices = graph.getVertices();
+    auto copyGraph = std::make_unique<ChcDirectedGraph>(graph);
+    std::vector<EId> loop = detectLoop(*copyGraph);
+    std::vector<WitnessInfo> loops;
+    while (loop.size() > 1) {
+        auto wtnsInfo = copyGraph->contractConnectedVertices(loop);
+        loops.push_back(wtnsInfo);
+        loop = detectLoop(*copyGraph);
+    }
+    return {std::move(copyGraph),
+            std::make_unique<NestedLoopTransformation::WitnessBackTranslator>(graph, *copyGraph, loops)};
+}
+
+VerificationResult NestedLoopTransformation::WitnessBackTranslator::translate(VerificationResult & result) {
+    if (result.getAnswer() == VerificationAnswer::UNSAFE) {
+        auto witness = translateErrorPath(result.getInvalidityWitness());
+        if (std::holds_alternative<NoWitness>(witness)) {
+            return {result.getAnswer(), std::get<NoWitness>(std::move(witness))};
+        }
+        return {VerificationAnswer::UNSAFE, std::get<InvalidityWitness>(std::move(witness))};
+    }
+    if (result.getAnswer() == VerificationAnswer::SAFE) {
+        auto witness = translateInvariant(result.getValidityWitness());
+        if (std::holds_alternative<NoWitness>(witness)) {
+            return {result.getAnswer(), std::get<NoWitness>(std::move(witness))};
+        }
+        return {VerificationAnswer::SAFE, std::get<ValidityWitness>(std::move(witness))};
+    }
+    return std::move(result);
+}
+
+NestedLoopTransformation::WitnessBackTranslator::ErrorOr<InvalidityWitness>
+NestedLoopTransformation::WitnessBackTranslator::translateErrorPath(InvalidityWitness wtns) {
+    // We need to get the CEX path, which will define the locations in the graph
+    Logic & logic = initialGraph.getLogic();
+    TimeMachine tm(logic);
+    auto errorPath = wtns.getDerivation();
+    std::vector<SymRef> pathVertices;
+    pathVertices.push_back(initialGraph.getEntry());
+    auto allVertices = initialGraph.getVertices();
+
+    // Compute model for the error path
+    auto model = [&]() {
+        SMTSolver solverWrapper(logic, SMTSolver::WitnessProduction::ONLY_MODEL);
+        for (std::size_t i = 1; i < errorPath.size(); ++i) {
+            PTRef fla = graph.getEdgeLabel(errorPath[i].clauseId);
+            fla = TimeMachine(logic).sendFlaThroughTime(fla, i);
+            solverWrapper.assertProp(fla);
+        }
+        auto res = solverWrapper.check();
+        if (res != SMTSolver::Answer::SAT) { throw std::logic_error("Error in computing model for the error path"); }
+        return solverWrapper.getModel();
+    }();
+
+    for (std::size_t i = 1; i < errorPath.size(); ++i) {
+        SymRef target = graph.getTarget(errorPath[i].clauseId);
+        auto loopNode = std::find_if(loopContractionInfos.begin(), loopContractionInfos.end(),
+                                     [&target](const WitnessInfo & x) { return x.loopVertex == target; });
+        if (loopNode != loopContractionInfos.end()) {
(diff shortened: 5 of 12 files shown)


src/engine/TPA.cc
```diff
@@ -46,6 +47,16 @@ VerificationResult TPAEngine::solve(ChcDirectedHyperGraph const & graph) {
     auto translator = std::move(transformationResult.second);
     if (transformedGraph->isNormalGraph()) {
         auto normalGraph = transformedGraph->toNormalGraph();
+        if (options.hasOption(Options::SIMPLIFY_NESTED)) {
```
blishko, line 50: Can you move this into the `solve` method that takes `DirectedChcGraph`?

blishko (commented):
Can you help we understand this by maybe providing some comments or pseudocode?

Also, can you build several benchmarks where you would compare the performance before and after?

src/transformers/NestedLoopTransformation.cc
```diff
@@ -0,0 +1,457 @@
...
+
+    verticesOnStack.erase(node.x);
+}
+
+std::vector<EId> NestedLoopTransformation::detectLoop(const ChcDirectedGraph & graph) {
+    std::vector<EId> loop;
+    if (graph.getVertices().size() <= 3) { return loop; }
+    auto graphRepresentation = AdjacencyListsGraphRepresentation::from(graph);
+    auto vertices = reversePostOrder(graph, graphRepresentation);
+    std::unordered_set<int> visitedVertices;
```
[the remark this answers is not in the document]
BritikovKI: It was supposed to be SymRef, but using it creates problems:
image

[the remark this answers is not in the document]
BritikovKI: I can go and try create default constructor for SymRef, but I'm not sure if it is an appropriate solution here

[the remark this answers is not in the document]
blishko: You just need to provide `SymRefHash` as additional template parameter.
`std::unordered_set<SymRef, SymRefHash>`.

You have to tell the set how it is supposed to hash your custom type.

src/engine/TPA.cc
BritikovKI: Updated

src/transformers/NestedLoopTransformation.cc
```diff
@@ -0,0 +1,457 @@
...
+    PositionVarMap argVars;
+    while (loop.size() > 1) {
+        simplifiedNodes.push_back(simplifyLoop(graph, loop, locationVars, argVars));
+        loop = detectLoop(graph);
+    }
+    return std::make_unique<NestedLoopTransformation::WitnessBackTranslator>(
+        std::move(originalGraph), graph, std::move(locationVars), std::move(argVars), simplifiedNodes);
+}
+
+void strongConnection(std::unordered_set<int> & visitedVertices, std::unordered_set<int> & verticesOnStack,
```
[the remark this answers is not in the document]
BritikovKI: Updated

src/transformers/NestedLoopTransformation.cc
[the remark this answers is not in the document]
BritikovKI: Updated

src/TransformationUtils.cc
```diff
@@ -88,57 +88,63 @@ std::unique_ptr<TransitionSystem> toTransitionSystem(ChcDirectedGraph const & gr
     return ts;
 }
 
-bool strongConnection(std::unordered_set<int> & visitedVertices, std::unordered_set<int> & verticesOnStack,
+// This function follows a Tarjan's strong connection detection algorithm:
+// https://en.wikipedia.org/wiki/Tarjan's_strongly_connected_components_algorithm
+// If some vertice is added on stack and it was already visited, then the strong connection is detected
+// It was updated to remember the nodes which participate in the loop (lines 123, 127)
```
blishko, line 94: The line numbers do not correspond anymore.

src/TransformationUtils.cc
```diff
@@ -88,57 +88,63 @@ std::unique_ptr<TransitionSystem> toTransitionSystem(ChcDirectedGraph const & gr
...
+void strongConnection(std::unordered_set<SymRef, SymRefHash> & visitedVertices,
```
blishko, line 95: Since you do not implement full Tarjan algorithm, but only detect one loop, maybe the function should have a different name?

I am thinking that you could create a `FindFirstLoop` class and move the implementation there.
Then the method could be just `visit` for visiting a node.

src/TransformationUtils.cc
```diff
@@ -88,57 +88,63 @@ std::unique_ptr<TransitionSystem> toTransitionSystem(ChcDirectedGraph const & gr
...
-    verticesOnStack.insert(node.x);
+                      SymRef node, std::vector<EId> & loop) {
+    visitedVertices.insert(node);
+    verticesOnStack.insert(node);
     auto const & outEdges = graphRepresentation.getOutgoingEdgesFor(node);
     if (size(outEdges) <= 1) {
-        verticesOnStack.erase(node.x);
-        return false;
+        verticesOnStack.erase(node);
+        return;
```
blishko, line 104: Why can you ignore nodes with one outgoing edge?

src/TransformationUtils.cc
```diff
@@ -88,57 +88,63 @@ std::unique_ptr<TransitionSystem> toTransitionSystem(ChcDirectedGraph const & gr
...
-                bool loopFound =
-                    strongConnection(visitedVertices, verticesOnStack, graphRepresentation, graph, nextVertex);
-                if (loopFound) { return true; }
-            } else if (verticesOnStack.find(nextVertex.x) != verticesOnStack.end()) {
-                return true;
+            if (visitedVertices.find(nextVertex) == visitedVertices.end()) {
+                strongConnection(visitedVertices, verticesOnStack, graphRepresentation, graph, nextVertex, loop);
+                if (!loop.empty() && graph.getTarget(eid) != graph.getTarget(loop[0]) &&
+                    graph.getTarget(eid) == graph.getSource(loop[loop.size() - 1])) {
+                    loop.push_back(eid);
```
blishko, line 114: Can you explain this condition?

src/TransformationUtils.cc
```diff
@@ -88,57 +88,63 @@ std::unique_ptr<TransitionSystem> toTransitionSystem(ChcDirectedGraph const & gr
...
-    std::unordered_set<int> verticesOnStack;
-
-    for (uint i = 1; i < vertices.size() - 1; i++) {
-        if (visitedVertices.find(vertices[i].x) == visitedVertices.end()) {
-            bool loop_detected =
-                strongConnection(visitedVertices, verticesOnStack, graphRepresentation, graph, vertices[i]);
-            if (loop_detected) { return true; }
+    std::unordered_set<SymRef, SymRefHash> visitedVertices;
+    std::unordered_set<SymRef, SymRefHash> verticesOnStack;
+    for (uint i = 1; i < vertices.size() - 1 && loop.size() == 0; i++) {
```
blishko, line 134: Can you move the condition `loop.size() == 0` to the body of the loop. Here it is somewhat hidden.

Actually, wouldn't it be sufficient to run `strongConnection` on the entry of the graph?

src/TransformationUtils.cc
BritikovKI: Because we simplify loop beforehand, so if there is just one out-edge it is selfloop, defacto this condition can be changed to `size(outEdges) == 0` and it would work the same

src/TransformationUtils.cc
BritikovKI: I check if the loop I found is actually loop:
graph.getTarget(eid) != graph.getTarget(loop[0]) &&
                    graph.getTarget(eid) == graph.getSource(loop[loop.size() - 1]
This checks if the target of the last edge of the loop and the source of the first edge is the same

src/TransformationUtils.cc
BritikovKI: Updated

src/TransformationUtils.cc
BritikovKI: Updated

src/TransformationUtils.cc
BritikovKI: Updated

src/transformers/NestedLoopTransformation.cc
```diff
@@ -0,0 +1,342 @@
...
+}
+
+SymRef NestedLoopTransformation::simplifyLoop(ChcDirectedGraph & graph, std::vector<EId> loop,
+                                              LocationVarMap & locationVars, PositionVarMap & argVars) {
+    Logic & logic = graph.getLogic();
+    TimeMachine timeMachine(logic);
+    TermUtils utils(logic);
+    std::vector<SymRef> vertices;
+    assert(graph.getSource(loop[0]) == graph.getTarget(loop[loop.size() - 1]));
+    assert(loop.size() >= 2);
```
blishko, line 38: Move asserts to the top of the method.

src/transformers/NestedLoopTransformation.cc
```diff
@@ -0,0 +1,342 @@
...
+    for (int i = loop.size() - 1; i >= 0; i--) {
+        vertices.push_back(graph.getSource(loop[i]));
+    }
+    // KB: creation of the location variables for nodes and argument variables (params of predicates)
+    std::vector<EId> allEdges = graph.getEdges();
```
blishko, line 43: Move this to the place where it is actually used.

blishko (commented):
Some further comments.
Still need to get to the core of the transformation...

src/graph/ChcGraph.h
```diff
@@ -141,7 +153,7 @@ class ChcDirectedGraph {
     ChcDirectedGraph reverse() const;
     DirectedEdge reverseEdge(DirectedEdge const & edge, TermUtils & utils) const;
 
-    LinearCanonicalPredicateRepresentation getPredicateRepresentation() const { return predicates; }
+    LinearCanonicalPredicateRepresentation & getPredicateRepresentation() { return predicates; }
```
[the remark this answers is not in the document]
BritikovKI: Bc when I merge vertices, I create a new vertice, and create a new predicate for this vertice

I need to somehow add a new predicate to the set of vertices

src/transformers/NestedLoopTransformation.cc
BritikovKI: Done

src/transformers/NestedLoopTransformation.cc
BritikovKI: Done

src/transformers/LoopTransformation.h
```diff
@@ -0,0 +1,107 @@
...
+        }
+    };
+
+    using LocationVarMap = std::unordered_map<SymRef, PTRef, SymRefHash>;
+    using PositionVarMap = std::unordered_map<VarPosition, PTRef, VarPositionHasher>;
+
+};
+
+
+namespace {
```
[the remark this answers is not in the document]
BritikovKI: If I remove the header I get a bunch of problems during compilation (due to the dependency tree)

src/graph/ChcGraph.h
```diff
@@ -171,11 +183,36 @@ class ChcDirectedGraph {
...
+
+    void updateEdgeTarget(EId edge, SymRef target) {
+        edges[edge].to  = target;
+    }
+
+    void removeEdge(EId edge){
+        const auto it = edges.find(edge);
+        if (it != edges.end())
+            edges.erase(it);
+    }
```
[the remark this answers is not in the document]
BritikovKI: Done

src/transformers/LoopTransformation.h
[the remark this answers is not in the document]
blishko: You should move the function definitions to `.cc` file.

src/graph/ChcGraph.cc
```diff
@@ -599,4 +599,194 @@ std::unique_ptr<ChcDirectedHyperGraph> ChcDirectedHyperGraph::makeEmpty(Logic &
...
+    // This is sound because we still force the right variables to be considered using the location variables.
+    vec<PTRef> components{sourceLocationVar, translatedConstraint, timeMachine.sendVarThroughTime(targetLocationVar, 1),
+                          updatedLocation};
+    return logic.mkAnd(std::move(components));
+}
+
+WitnessInfo ChcDirectedGraph::contractConnectedVertexes(std::vector<EId> edges) {
+    assert(edges.size() >= 2);
+    assert(getSource(edges[0]) == getTarget(edges[edges.size() - 1]));
+    TimeMachine timeMachine(logic);
```
BritikovKI, line 657: Mb it's better to create `timeMachine` and `termUtils` once per graph during construction then call constructor again and again?

There are 23 constructor calls to termUtils just in this file!

blishko: `TimeMachine` and `TermUtils` are trivial objects, they just store a reference to `Logic`.
It is OK to create them on demand.

BritikovKI:
Ok, I think generally the code was moved and main upds are done

Pls take a look to understand if it's Ok architecturally or if some additional cleanup is needed!

src/graph/ChcGraph.cc
```diff
@@ -6,6 +6,8 @@
 
 #include "ChcGraph.h"
 
+#include "../transformers/CommonUtils.h"
+#include "../TransformationUtils.h"
```
blishko, line 10:
```suggestion
#include "transformers/CommonUtils.h"
#include "TransformationUtils.h"
```

This should be OK, no?

src/graph/ChcGraph.cc
```diff
@@ -599,4 +601,126 @@ std::unique_ptr<ChcDirectedHyperGraph> ChcDirectedHyperGraph::makeEmpty(Logic &
...
+WitnessInfo ChcDirectedGraph::contractConnectedVertexes(std::vector<EId> edges) {
+    assert(edges.size() >= 2);
+    assert(getSource(edges[0]) == getTarget(edges[edges.size() - 1]));
+    TimeMachine timeMachine(logic);
+    TermUtils utils(logic);
+    std::vector<SymRef> vertices;
+    for (int i = edges.size() - 1; i >= 0; i--) {
+        assert((i > 0 && getSource(edges[i]) == getTarget(edges[i-1])) || i == 0);
+        vertices.push_back(getSource(edges[i]));
+    }
```
blishko, line 615:
```suggestion
    assert(std::adjacent_find(edges.begin(), edges.end(), [&](auto first, auto second) { return getTarget(first) != getTarget(second); }) == edges.end());
    for (int i = edges.size() - 1; i >= 0; i--) {
        vertices.push_back(getSource(edges[i]));
    }
```

src/graph/ChcGraph.cc
```diff
@@ -599,4 +601,126 @@ std::unique_ptr<ChcDirectedHyperGraph> ChcDirectedHyperGraph::makeEmpty(Logic &
...
+        for (auto && entry : locationVars) {
+            ret.push_back(entry.second);
+        }
+        for (auto && entry : argVars) {
+            ret.push_back(entry.second);
+        }
+        return ret;
+    }();
+
+    vec<SRef> args;
```
blishko, line 663:
```suggestion
    vec<SRef> argSorts;
```

src/graph/ChcGraph.cc
```diff
@@ -599,4 +601,126 @@ std::unique_ptr<ChcDirectedHyperGraph> ChcDirectedHyperGraph::makeEmpty(Logic &
...
+
+    SymRef newRef = logic.declareFun(name, logic.getSort_bool(), args);
+    predicates.addRepresentation(newRef, std::move(vars));
+
+    // KB: Changing input and output edges to and from the loop in the graph with new state and loc variables
+    //     Redirecting edges to the newly created vertice
+    allEdges = getEdges();
+    for (auto edge : allEdges) {
+        if (std::find(vertices.begin(), vertices.end(), getSource(edge)) != vertices.end()) {
+            PTRef badStates = locationVars.at(getSource(edge));
```
blishko, line 684: `badStates` is not the right name here.

src/graph/ChcGraph.cc
```diff
@@ -599,4 +601,126 @@ std::unique_ptr<ChcDirectedHyperGraph> ChcDirectedHyperGraph::makeEmpty(Logic &
...
+                subMap.insert(std::make_pair(oldEdgeVars.stateVars[i], argVars[{getSource(edge), i}]));
+            }
+            updateEdgeSource(edge, newRef);
+            PTRef label = utils.varSubstitute(logic.mkAnd(getEdgeLabel(edge), badStates), subMap);
+
+            updateEdgeLabel(edge, InterpretedFla{label});
+        }
+        if (std::find(vertices.begin(), vertices.end(), getTarget(edge)) != vertices.end()) {
+
+            PTRef initialStates = [&]() -> PTRef {
```
blishko, line 697: `initialStates` is not the right name here.

src/graph/ChcGraph.cc
```diff
@@ -599,4 +601,126 @@ std::unique_ptr<ChcDirectedHyperGraph> ChcDirectedHyperGraph::makeEmpty(Logic &
...
+                vec<PTRef> negatedLocations;
+                negatedLocations.capacity(locationVars.size());
+                for (auto && entry : locationVars) {
+                    if (std::get<0>(entry) != getTarget(edge)) {
```
blishko, line 701:
```suggestion
                    if (entry.first != getTarget(edge)) {
```

src/graph/ChcGraph.cc
```diff
@@ -599,4 +601,126 @@ std::unique_ptr<ChcDirectedHyperGraph> ChcDirectedHyperGraph::makeEmpty(Logic &
...
+            PTRef label = utils.varSubstitute(getEdgeLabel(edge), subMap);
+
+            updateEdgeLabel(edge, InterpretedFla{rewriteMaxArityClassic(
+                                            logic, logic.mkAnd(logic.mkAnd(initialStates, destination), label))});
+        }
+    }
+
+    // KB: Creating new self-looping edge (which represents all of the transitions in the outer loop)
+    newEdge(newRef, newRef, InterpretedFla{transitionRelation});
+    return {newRef, locationVars, argVars};
```
blishko, line 725: I am still concerned about the case where there are additional edges between the nodes of the loop.
You will end up with multiple self-loops on the new node, which I think is not want you want.
You want to have single self-loop in the end, no?

src/graph/ChcGraph.h
```diff
@@ -141,6 +165,20 @@ class ChcDirectedGraph {
     ChcDirectedGraph reverse() const;
     DirectedEdge reverseEdge(DirectedEdge const & edge, TermUtils & utils) const;
 
+    ChcDirectedGraph copy(const ChcDirectedGraph& graph)
```
blishko, line 168: Why do you need special `copy` method? Why is not the default copy constructor good enough?

src/graph/ChcGraph.h
```diff
@@ -187,13 +225,32 @@ class ChcDirectedGraph {
         }
     }
 
+
```
blishko, line 228:
```suggestion
```

src/transformers/NestedLoopTransformation.cc
```diff
@@ -0,0 +1,213 @@
...
+    auto errorPath = wtns.getDerivation();
+    std::vector<SymRef> pathVertices;
+    pathVertices.push_back(initialGraph.getEntry());
+    auto newVertices = graph.getVertices();
+    auto allVertices = initialGraph.getVertices();
+
+    // Compute model for the error path
+    auto model = [&]() {
+        SMTSolver solverWrapper(logic, SMTSolver::WitnessProduction::ONLY_MODEL);
+        auto & solver = solverWrapper.getCoreSolver();
```
blishko, line 60: There is no longer a need to access the wrapped solver.
You can achieve what you need with the interface of `SMTSolver` now.

src/transformers/NestedLoopTransformation.cc
```diff
@@ -0,0 +1,213 @@
...
+    auto oldVertices = initialGraph.getVertices();
+    TermUtils utils(logic);
+    TermUtils::substitutions_map substitutions;
+    for (auto vertex : oldVertices) {
+        if (vertex == graph.getEntry()) { continue; }
+        for (auto loop : loops) {
+            try {
+                PTRef locationVar = loop.locations.at(vertex);
+                substitutions.insert({timeMachine.getUnversioned(locationVar), logic.getTerm_false()});
+            } catch (const std::out_of_range & ex) { continue; }
```
blishko, line 128: Please don't use exceptions to check if some element is in a map.

src/transformers/NestedLoopTransformation.cc
```diff
@@ -0,0 +1,213 @@
...
+    for (auto v : loops) {
+        PTRef unversionedPredicate =
+            TimeMachine(logic).versionedFormulaToUnversioned(graph.getStateVersion(v.loopVertex));
+        PTRef mergedInv = wtns.getDefinitions().at(unversionedPredicate);
+        for (auto vertex : oldVertices) {
+            if (vertex == initialGraph.getEntry() or vertex == initialGraph.getExit()) { continue; }
+            PTRef locationVar;
+            try {
+                locationVar = timeMachine.getUnversioned(v.locations.at(vertex));
+            } catch (const std::out_of_range & ex) { continue; }
```
blishko, line 146: Please don't use exceptions to check if some element is in a map.

blishko (commented):
Can you please also add tests for the new functionality?

src/transformers/NestedLoopTransformation.cc
BritikovKI: Fixed

src/graph/ChcGraph.h
BritikovKI: Fixed

src/graph/ChcGraph.cc
BritikovKI: Fixed

src/graph/ChcGraph.cc
BritikovKI: Fixed

src/graph/ChcGraph.cc
BritikovKI: Fixed

src/graph/ChcGraph.cc
BritikovKI: Fixed

src/graph/ChcGraph.cc
BritikovKI: Fixed

src/graph/ChcGraph.cc
BritikovKI: This should be taken care of at the lines 642-647
All of the possible edges between the nodes inside the loop are removed

blishko: OK, I see. I agree that that should work correctly.

src/graph/ChcGraph.h
BritikovKI: Well, now it worked, Idk what it was before 😅 
Thanks!

BritikovKI:
> Can you please also add tests for the new functionality?

Done, added a couple of tests on merging the loops!

src/engine/TPA.cc
```diff
@@ -73,8 +74,17 @@ VerificationResult TPAEngine::solve(const ChcDirectedGraph & graph) {
                 assert(false);
                 throw std::logic_error("Unreachable!");
         }
-    } else if (isTransitionSystemDAG(graph) && not options.hasOption(Options::FORCE_TS)) {
-        return solveTransitionSystemGraph(graph);
+    } else if ((isTransitionSystemDAG(graph) && not options.hasOption(Options::FORCE_TS)) ||
+               options.hasOption(Options::SIMPLIFY_NESTED)) {
+        if (options.hasOption(Options::SIMPLIFY_NESTED)) {
+            NestedLoopTransformation transformation;
+            auto normalGraph = graph;
```
blishko, line 81: You don't need a copy here, no?

src/engine/TPA.cc
```diff
@@ -73,8 +74,17 @@ VerificationResult TPAEngine::solve(const ChcDirectedGraph & graph) {
...
+            auto preTranslator = transformation.transform(normalGraph);
+            auto res = solveTransitionSystemGraph(normalGraph);
```
blishko, line 83: Add an `assert` that this `isTransitionSystemGraph` now.

src/graph/ChcGraph.cc
```diff
@@ -599,4 +601,126 @@ std::unique_ptr<ChcDirectedHyperGraph> ChcDirectedHyperGraph::makeEmpty(Logic &
         std::move(predicateRepresentation),
         logic
         );
+}
+
+WitnessInfo ChcDirectedGraph::contractConnectedVertexes(std::vector<EId> edges) {
```
blishko, line 606:
```suggestion
WitnessInfo ChcDirectedGraph::contractConnectedVertices(std::vector<EId> edges) {
```

Also, any reason why this cannot take `const &` or `&&` instead of a copy of the loop?

src/graph/ChcGraph.cc
```diff
@@ -599,4 +601,126 @@ std::unique_ptr<ChcDirectedHyperGraph> ChcDirectedHyperGraph::makeEmpty(Logic &
...
+    }();
+
+    vec<SRef> argSorts;
+    std::vector<PTRef> vars;
+    for (auto arg : stateVars) {
+        argSorts.push(logic.getSortRef(arg));
+        vars.push_back(timeMachine.getUnversioned(arg));
+    }
+
+    PTRef transitionRelation = logic.mkOr(std::move(transitionRelationComponent));
```
blishko, line 670: Move up just after you compute all the components.

src/graph/ChcGraph.cc
```diff
@@ -599,4 +601,126 @@ std::unique_ptr<ChcDirectedHyperGraph> ChcDirectedHyperGraph::makeEmpty(Logic &
...
+            std::unordered_map<PTRef, PTRef, PTRefHash> subMap;
+            for (uint i = 0; i < oldEdgeVars.nextStateVars.size(); i++) {
+                subMap.insert(std::make_pair(oldEdgeVars.nextStateVars[i],
+                                             timeMachine.sendVarThroughTime(argVars[{getTarget(edge), i}], 1)));
+            }
+            updateEdgeTarget(edge, newRef);
+            PTRef label = utils.varSubstitute(getEdgeLabel(edge), subMap);
+
+            updateEdgeLabel(edge, InterpretedFla{rewriteMaxArityClassic(
+                                            logic, logic.mkAnd(logic.mkAnd(newTargetLoc, destination), label))});
```
blishko, line 719:
```suggestion
            updateEdgeLabel(edge, InterpretedFla{logic.mkAnd({newTargetLoc, destination, label})});
```
This should be enough, no?

src/transformers/NestedLoopTransformation.h
```diff
@@ -0,0 +1,68 @@
...
+ * values for location variables. This may still leave some undesired variables in the invariant. We currently make
+ * best effort to eliminate these variables by simplifying the formula.
+ */
+class NestedLoopTransformation {
+
+public:
+    class WitnessBackTranslator {
+        ChcDirectedGraph initialGraph;
+        ChcDirectedGraph const & graph;
+        std::vector<WitnessInfo> loops;
```
blishko, line 46:
```suggestion
        std::vector<WitnessInfo> loopContractionInfos;
```

test/test_Transformers.cc
```diff
@@ -802,4 +805,173 @@ TEST_F(Transformer_New_Test, test_SummarizerAndUnreachable_WitnessBacktranslatio
...
+    for (auto const & clause : clauses) { system.addClause(clause); }
+
+    Logic & logic = *this->logic;
+    auto normalizedSystem = Normalizer(logic).normalize(system);
+    auto hyperGraph = ChcGraphBuilder(logic).buildGraph(normalizedSystem);
+    auto originalGraph = *hyperGraph;
+    auto transformation = Transformations::towardsTransitionSystems();
+    auto [transformedGraph, translator] = transformation.transform(std::move(hyperGraph));
+    NestedLoopTransformation NLTransformation;
+    auto propperGraph = transformedGraph->toNormalGraph();
```
blishko, line 881:
```suggestion
    auto properGraph = transformedGraph->toNormalGraph();
```

test/test_Transformers.cc
```diff
@@ -802,4 +805,173 @@ TEST_F(Transformer_New_Test, test_SummarizerAndUnreachable_WitnessBacktranslatio
...
+            ChcBody{{logic->mkAnd(logic->mkGeq(y, ten), logic->mkLeq(x, zero))}, {UninterpretedPredicate{currentS1}}}
+        }};
+
+    for (auto const & clause : clauses) { system.addClause(clause); }
+
+    Logic & logic = *this->logic;
+    auto normalizedSystem = Normalizer(logic).normalize(system);
+    auto hyperGraph = ChcGraphBuilder(logic).buildGraph(normalizedSystem);
+    auto originalGraph = *hyperGraph;
+    auto transformation = Transformations::towardsTransitionSystems();
```
blishko, line 878: Is this needed?

blishko (commented):
It looks good! Just some final cleanup!

src/engine/TPA.cc
BritikovKI: Nope, I need it bc `graph` is a const, and I'm changing `normalGraph` inside of the `transform` function

BritikovKI: I also need a const version inside of `transform` to store in the witness

blishko: You can change `transform` to take `const &`, no?
Since you are making a copy inside anyway, you can then operate on the copy.

test/test_Transformers.cc
BritikovKI: Removed

test/test_Transformers.cc
BritikovKI: Fixed

src/transformers/NestedLoopTransformation.h
BritikovKI: Fixed

src/graph/ChcGraph.cc
BritikovKI: Fixed

src/graph/ChcGraph.cc
BritikovKI: Fixed

src/graph/ChcGraph.cc
BritikovKI: Fixed

src/engine/TPA.cc
BritikovKI: Fixed

src/engine/TPA.cc
BritikovKI: Ok, didn't want to return modified copy, will do

BritikovKI: Fixed

src/transformers/NestedLoopTransformation.cc
```diff
@@ -0,0 +1,214 @@
...
+    auto vertices = graph.getVertices();
+    auto copyGraph = graph;
+    std::vector<EId> loop = detectLoop(copyGraph);
+    std::vector<WitnessInfo> loops;
+    while (loop.size() > 1) {
+        auto wtnsInfo = copyGraph.contractConnectedVertices(loop);
+        loops.push_back(wtnsInfo);
+        loop = detectLoop(copyGraph);
+    }
+    return std::make_unique<NestedLoopTransformation::WitnessBackTranslator>(graph, std::move(copyGraph), loops);
```
blishko, line 25: That's not what I meant.
The backtranslator does not need to know about the new graph, no?
If you look how other transformations work, they return a pair of unique pointers. One to the transformed graph, one to the backtranslator.
We should use the same design here.

src/TransformationUtils.cc
```diff
@@ -134,6 +134,7 @@ std::vector<EId> detectLoop(const ChcDirectedGraph & graph) {
     std::unordered_set<SymRef, SymRefHash> visitedVertices;
     std::unordered_set<SymRef, SymRefHash> verticesOnStack;
     visit(visitedVertices, verticesOnStack, graphRepresentation, graph, graph.getEntry(), loop);
+    reverse(loop.begin(), loop.end());
```
blishko, line 137: Why do you need to reverse the loop now?

blishko: Also, should be `std::reverse`, no?

BritikovKI: Main reason is I want to have a representation of the loop in which it is present in the current active path, but before reverse it is in the reversed order

src/transformers/NestedLoopTransformation.cc
BritikovKI: Backtranslator needs to know about the new graph though.

It is being reused to produce model from the invalidity Witness, to navigate the new graph and based on it to transform. witness into the new representation

Or do you mean we should pass newly produced graph to the witness translation as one of the arguments, but not as a class member?

blishko: OK, I see.
Then I suggest that the Backtranslator has a shared pointer to the transformed graph and you also return a shared pointer to the same graph from `translate`.

src/TransformationUtils.cc
blishko: OK. Just make it `std::reverse`.

src/transformers/NestedLoopTransformation.cc
BritikovKI: Fixed

src/TransformationUtils.cc
BritikovKI: Fixed

blishko (approved):
LGTM!

blishko:
> Can you please rebase?

And squash the commits? I think this can be a single commit, no?

PR #89, opened 2024-10-25
Loop detection rework

BritikovKI:
Reworked loop detection to return found loops.

This is needed for the future work on the nested loops!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
 build
-build-static
 
 ## idea
 cmake-build-*
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 build
+build-static
 
 ## idea
 cmake-build-*
--- a/src/TransformationUtils.cc
+++ b/src/TransformationUtils.cc
@@ -88,10 +88,11 @@ std::unique_ptr<TransitionSystem> toTransitionSystem(ChcDirectedGraph const & gr
     return ts;
 }
 
-// This function follows a Tarjan's strong connection detection algorithm:
+namespace {
+// This function follows a Tarjan's strongly connected component detection algorithm:
 // https://en.wikipedia.org/wiki/Tarjan's_strongly_connected_components_algorithm
 // If some vertice is added on stack and it was already visited, then the strong connection is detected
-// It was updated to remember the nodes which participate in the loop (lines 113, 119)
+// It was updated to remember the nodes which participate in the loop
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -1,4 +1,5 @@
 build
+build-static
```
blishko, line 2: Let's not do this in this PR!

src/TransformationUtils.cc
```diff
@@ -88,57 +88,60 @@ std::unique_ptr<TransitionSystem> toTransitionSystem(ChcDirectedGraph const & gr
...
 
-bool strongConnection(std::unordered_set<int> & visitedVertices, std::unordered_set<int> & verticesOnStack,
-                      AdjacencyListsGraphRepresentation const & graphRepresentation, ChcDirectedGraph const & graph,
-                      SymRef node) {
-    visitedVertices.insert(node.x);
-    verticesOnStack.insert(node.x);
+// This function follows a Tarjan's strong connection detection algorithm:
+// https://en.wikipedia.org/wiki/Tarjan's_strongly_connected_components_algorithm
+// If some vertice is added on stack and it was already visited, then the strong connection is detected
+// It was updated to remember the nodes which participate in the loop (lines 113, 119)
```
blishko, line 94: Are these line numbers correct?
Line 119 is `break`.
We probably should not refer to line numbers in here, it can very easily become incorrect.
Also, please replace `strong connection` with `strongly connected component`.

src/TransformationUtils.cc
```diff
@@ -88,57 +88,60 @@ std::unique_ptr<TransitionSystem> toTransitionSystem(ChcDirectedGraph const & gr
...
+// Algorithm terminates as soon as it can find the first loop
+void visit(std::unordered_set<SymRef, SymRefHash> & visitedVertices,
```
blishko, line 96: Please put this helper function inside an anonymous namespace.

blishko (commented):
Looks good, just a small cleanup is needed!

.gitignore
BritikovKI: 💔

src/TransformationUtils.cc
BritikovKI: Done

.gitignore
BritikovKI: Done, though a bit sad about it 😞

blishko (approved):
LGTM!

PR #90, opened 2024-10-25
Loop transformation refactoring

BritikovKI:
Some parts of loop transformation are moved to the Utils.
It is done to generalise the merging of the loops/vertices in the graph


src/transformers/CommonUtils.h
```diff
@@ -9,6 +9,7 @@
 
 #include "graph/ChcGraph.h"
 #include "Witnesses.h"
+#include "TransformationUtils.h"
```
blishko, line 12: Why do you need this include?

BritikovKI: This one is needed for `getVariablesFromEdge` function which is used in EdgeTranslate

BritikovKI:
Btw, should we mb add CommonUtils to clang-formatter?
Wdyt?

src/transformers/CommonUtils.h
blishko: Then it should be included in `.cc` file, not in the header.

blishko:
> Btw, should we mb add CommonUtils to clang-formatter? Wdyt?

Let's not do that now, I am thinking of merging `CommonUtils` and `TransformationUtils`.

BritikovKI:
Ok, updated, I think if we want to merge CommonUtils and TransformationUtils it should be done separately...

blishko:
> Ok, updated, I think if we want to merge CommonUtils and TransformationUtils it should be done separately...

Indeed.

blishko (approved):
LGTM!

PR #159, opened 2025-12-15
Interpolation-based Transition Invariants for Termination

BritikovKI:
Extension of the safety-based nontermination detection with the parallel construction of transition invariants for transition relation reusing detected terminating traces by the safety-based nontermination.

These transition invariants are used to attempt proving termination during the nontermination detection procedure.

### DIFF
--- a/src/termination/ReachabilityNonterm.cc
+++ b/src/termination/ReachabilityNonterm.cc
@@ -853,7 +853,7 @@ std::tuple<ReachabilityNonterm::Answer, PTRef> ReachabilityNonterm::analyzeTS(PT
                 PTRef noncoveredStates = QuantifierElimination(logic).keepOnly(
                     logic.mkAnd({logic.mkOr(trInv, id), TimeMachine(logic).sendFlaThroughTime(temp_tr, 1),
                                  logic.mkNot(shiftOnlyNextVars(trInv, vars, logic))}), vars);
-                // std::cout << "Noncovered: " << logic.pp(noncoveredStates) << std::endl;
+                std::cout << "Noncovered: " << logic.pp(noncoveredStates) << std::endl;
 
                  // We check if the states that are not covered by TrInv are reachable
                 auto graph = constructHyperGraph(init, transition,
@@ -902,7 +902,7 @@ std::tuple<ReachabilityNonterm::Answer, PTRef> ReachabilityNonterm::analyzeTS(PT
                         //     auto moved = TimeMachine(logic).sendVarThroughTime(var, num_non);
                         //     state.push(logic.mkEq(moved, m->evaluate(moved)));
                         // }
-                        // reached = TimeMachine(logic).sendFlaThroughTime(logic.mkAnd(state), -1);
+                         // reached = TimeMachine(logic).sendFlaThroughTime(logic.mkAnd(state), -1);
 
                         // reached = TermUtils(logic).getTopLevelDisjuncts(dnfize(TermUtils(logic).simplifyMax(TimeMachine(logic).sendFlaThroughTime(ModelBasedProjection(logic).keepOnly(transitions, last_vars, *smt_checker.getModel()), -num_non)), logic))[0];
                         reached =TermUtils(logic).simplifyMax(TimeMachine(logic).sendFlaThroughTime(ModelBasedProjection(logic).keepOnly(transitions, last_vars, *smt_checker.getModel()), -num_non));
@@ -910,15 +910,15 @@ std::tuple<ReachabilityNonterm::Answer, PTRef> ReachabilityNonterm::analyzeTS(PT
                     }
 
                     assert(reached != logic.getTerm_false());
-                    // std::cout << "init: " << logic.pp(init) << std::endl;
-                    // std::cout << "Reached: " << logic.pp(reached) << std::endl;
-                    // std::cout << "TrInv: " << logic.pp(logic.mkNot(trInv)) << std::endl;
-                    // std::cout << "Sink: " << logic.pp(logic.mkNot(noncoveredStates)) << std::endl;
+                    std::cout << "init: " << logic.pp(init) << std::endl;
+                    std::cout << "Reached: " << logic.pp(reached) << std::endl;
+                    std::cout << "TrInv: " << logic.pp(logic.mkNot(trInv)) << std::endl;
+                    std::cout << "Sink: " << logic.pp(sink) << std::endl;
                     // Algorithm checks if reachable states are terminating
-                    // std::cout<<"Deeper\n";
+                    std::cout<<"Deeper\n";
                     auto [answer, subinv] =
                         analyzeTS(reached, transition,  logic.mkNot(noncoveredStates), witnesses, logic, vars, DETERMINISTIC_TRANSITION);
-                    // std::cout<<"Higher\n";
+                    std::cout<<"Higher\n";
                     // TODO: If it terminates for noncoveredStates, then it terminates for all states
                     if (answer == Answer::YES) {
                         smt_checker.resetSolver();
@@ -1049,7 +1049,7 @@ ReachabilityNonterm::Answer ReachabilityNonterm::run(TransitionSystem const & ts
     Options witnesses = options;
     witnesses.addOption(options.COMPUTE_WITNESS, "true");
     bool DETERMINISTIC_TRANSITION = determinismCheck(transition, logic, vars);
-    auto [answer, trInvOrRecurringSet] = analyzeTS(init, transition, sink, witnesses, logic, vars, DETERMINISTIC_TRANSITION);
+    auto [answer, trInvOrRecurringSet] = analyzeTS(init, transition, dnfize(sink,logic), witnesses, logic, vars, DETERMINISTIC_TRANSITION);
     return answer;
 }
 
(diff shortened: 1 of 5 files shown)


src/termination/ReachabilityNonterm.cc
```diff
@@ -17,6 +17,321 @@
...
+void getCoeffs(ArithLogic & logic, std::vector<PTRef> & coefs, PTRef formula) {
+    assert(logic.isLeq(formula));
+    auto it = logic.getPterm(formula).begin();
+    assert(logic.getPterm(formula).size() == 2);
+    unrollAtom(logic, coefs, it[0], false);
+    unrollAtom(logic, coefs, it[1], true);
+}
+
+// Function to turn everything in <= formulas
+void lequalize(PTRef conjunct, vec<PTRef> & leqs, vec<PTRef> & bools, ArithLogic & logic) {
```
blishko: Doesn't Logic already do this for you?

src/termination/ReachabilityNonterm.cc
```diff
@@ -62,25 +377,149 @@ std::unique_ptr<ChcDirectedHyperGraph> constructHyperGraph(PTRef const init, PTR
...
+            {TimeMachine(logic).sendVarThroughTime(vars[i], 1), TimeMachine(logic).sendVarThroughTime(vars[i], 2)});
+        neq.push(logic.mkNot(logic.mkEq(TimeMachine(logic).sendVarThroughTime(vars[i], 1),
+                                        TimeMachine(logic).sendVarThroughTime(vars[i], 2))));
+    }
+    PTRef newTransition = TermUtils(logic).varSubstitute(transition, detSubstitutions);
+    // Tr(x,x') /\ Tr(x, x'') /\ ! x' = x''
+    detChecker.assertProp(logic.mkAnd({transition, newTransition, logic.mkOr(neq)}));
+
+    if (detChecker.check() == SMTSolver::Answer::UNSAT) {
+        std::cout << "DETERMINISTIC;" << std::endl;
```
blishko: Remove debug output

blishko (commented):
A few comments from first glance.
I'll take a proper look at the algorithm later.

src/termination/ReachabilityNonterm.cc
```diff
@@ -62,104 +275,224 @@ std::unique_ptr<ChcDirectedHyperGraph> constructHyperGraph(PTRef const init, PTR
...
+        extractActiveLiterals(TermUtils(logic).simplifyMax(logic.mkNot(formula)), logic, model, activeLiterals, solver,
+                              !reverse);
+    } else {
+        activeLiterals.push(reverse ? logic.mkNot(formula) : formula);
+    }
+}
+
+// Converts formula into disjunctive normal form
+// bound sets a maximal number of disjuncts in DNF if bound != 0
+PTRef toDNF(PTRef formula, Logic & logic, int bound = 0) {
```
blishko: If bound is set, then this gives you only underapproximation, no?
That should be clear from the comment and from the name of the function.

src/termination/ReachabilityNonterm.cc
```diff
@@ -62,104 +275,224 @@ std::unique_ptr<ChcDirectedHyperGraph> constructHyperGraph(PTRef const init, PTR
...
+}
+
+// This function extracts well-founded disjuncts from the interpolant
+vec<PTRef> extractWellFoundedCandidates(PTRef itp, PTRef sink, ArithLogic & logic, const std::vector<PTRef> & vars) {
+    TermUtils utils(logic);
+    SMTSolver smt_solver(logic, SMTSolver::WitnessProduction::NONE);
+
+    auto sink_disjuncts =
+        utils.getTopLevelDisjuncts(toDNF(utils.simplifyMax(normalize(logic.mkNot(sink), logic)), logic));
+    auto candidates = utils.getTopLevelDisjuncts(toDNF(utils.simplifyMax(normalize(itp, logic)), logic, 500));
```
blishko: Can you use name parameter instead of a magic constant number?

src/termination/ReachabilityNonterm.cc
```diff
@@ -62,104 +275,224 @@ std::unique_ptr<ChcDirectedHyperGraph> constructHyperGraph(PTRef const init, PTR
...
+        } else {
+            for (auto sink_cand : sink_disjuncts) {
+                PTRef conjunction = utils.simplifyMax(logic.mkAnd(sink_cand, cand));
+                smt_solver.resetSolver();
+                smt_solver.assertProp(conjunction);
+                if (smt_solver.check() == SMTSolver::Answer::SAT && checkWellFounded(conjunction, logic, vars)) {
+                    strictCandidates.push(conjunction);
+                }
+            }
+        }
```
blishko, line 364: Have you considered caching results of well-foundedness checks?
I would expect that it happens that you check the same candidate multiple times, no?

src/termination/ReachabilityNonterm.cc
```diff
@@ -62,104 +275,224 @@ std::unique_ptr<ChcDirectedHyperGraph> constructHyperGraph(PTRef const init, PTR
...
+    }
+    return strictCandidates;
+}
+
+// Function evaluates if Tr(x,x') determenistic
+bool determinismCheck(const PTRef & transition, Logic & logic, const std::vector<PTRef> & vars) {
```
blishko:
```suggestion
bool determinismCheck(PTRef const transition, Logic & logic, std::vector<PTRef> const & vars) {
```

src/termination/ReachabilityNonterm.cc
```diff
@@ -62,104 +275,224 @@ std::unique_ptr<ChcDirectedHyperGraph> constructHyperGraph(PTRef const init, PTR
...
+    for (auto var : vars) {
+        PTRef curr = TimeMachine(logic).sendVarThroughTime(var, 0);
+        PTRef next = TimeMachine(logic).sendVarThroughTime(var, 1);
+        eq_vars.push(logic.mkEq(curr, next));
+    }
+    return logic.mkAnd(eq_vars);
+}
+
+PTRef constructTransitionInvariantCandidates(PTRef init, PTRef transition, PTRef sink, int depth, Logic & logic,
+                                             const std::vector<PTRef> & vars) {
```
blishko:
```suggestion
                                            std::vector<PTRef> const & vars) {
```

src/termination/ReachabilityNonterm.cc
```diff
@@ -62,104 +275,224 @@ std::unique_ptr<ChcDirectedHyperGraph> constructHyperGraph(PTRef const init, PTR
...
+    checked_states.push_back(TimeMachine(logic).sendFlaThroughTime(sink, depth));
+    // sink is updated, representing states that are guaranteed to reach termination
+    PTRef terminating_states = logic.mkOr(checked_states);
+
+    // Itp(Tr /\ ... /\ Tr, Init /\ not TerminatingStates) should be UNSAT: by construction,
+    // `terminating_states` already covers everything reachable from `init` via the trace.
+    SMTSolver smt_solver(logic, SMTSolver::WitnessProduction::ONLY_INTERPOLANTS);
+    smt_solver.getConfig().setSimplifyInterpolant(4);
+    smt_solver.assertProp(trace);
+    smt_solver.push();
```
blishko: I don't see a reason for this `push`.

src/termination/ReachabilityNonterm.cc
```diff
@@ -62,104 +275,224 @@ std::unique_ptr<ChcDirectedHyperGraph> constructHyperGraph(PTRef const init, PTR
...
-            }
-            if (!detected) {
-                // If all transitions were determenistic, we block the initial states that lead to the termination
-                init = logic.mkAnd(init, logic.mkNot(Result));
-            }
+            PTRef trace = logic.mkAnd(formulas);
+            PTRef originalTransition = transition;
+            std::tie(init, transition) = blockDeterministicPrefix(init, transition, sink, trace, num, logic);
 
+            SMTSolver SMTsolver(logic, SMTSolver::WitnessProduction::NONE);
```
blishko:
```suggestion
            SMTSolver smtSolver(logic, SMTSolver::WitnessProduction::NONE);
```
Please use a different name for the variable than the type name.

blishko:
Can you squash the commits? It seems to me this could be a single commit. Or are there intermediate steps you want to preserve?

src/termination/ReachabilityNonterm.cc
BritikovKI: Done

src/termination/ReachabilityNonterm.cc
BritikovKI: Done

src/termination/ReachabilityNonterm.cc
BritikovKI: I want to slit A and B parts of interpolation into 2 different frames
Hypothetically it should operate the same even with a single frame, however, as I noticed, not using `push` loses around 10-15 instances for some reason

src/termination/ReachabilityNonterm.cc
BritikovKI: Fixed

src/termination/ReachabilityNonterm.cc
BritikovKI: This is a good idea, leads to slight improvement of the performance because indeed there are a lot of same candidates

src/termination/ReachabilityNonterm.cc
BritikovKI: Done

src/termination/ReachabilityNonterm.cc
BritikovKI: Fixed, though not sure about the function name, because it still does full DNFization with bound = 0

blishko (approved):
Let's merge this version and we can clean it up a bit later!

PR #165, opened 2026-03-12
Fix access out of bounds in `toTransitionSystem`

BritikovKI:
Anna detected a bug, which was appearing in the debug mode run on the Linux Endeavour, with both gcc and clang compilers.
The problem is, that when `toTransitionSystem` is called for Termination problem, it does not have 3 vertices, just 2.
Therefore checking `target == vertices[2]` is undefined behaviour.


blishko (approved):
The whole function would benefit would a rewrite, but this change is minimal and fixes the problem.

PR #166, opened 2026-04-20
added ic3ia engine to golem for transition systems

gianredondi:
## Summary

This PR introduces [IC3IA](https://es-static.fbk.eu/people/griggio/ic3ia/index.html) as an engine for transition systems in Golem.

This is a CEGAR-based approach for safety verification: predicates are mined from the initial formula and the target safety properties → predicate labels are introduced → IC3 runs on abstract system → if safe, done; if unsafe, spurious check and new predicates might be added via interpolation → loop restarts.

Includes simple options for different generalization in IC3, and refinement strategies.

Co-authored with Claude Sonnet 4.6, mainly to navigate Golem and OpenSMT API.

## Usage

```bash
golem -l QF_LIA -e ic3ia input.smt2

### DIFF
--- a/src/engine/IC3IA.h
+++ b/src/engine/IC3IA.h
@@ -46,8 +46,6 @@ public:
 
 private:
     // --- IC3 core types ---
-
-    // A cube is a conjunction of literals over the current state variables.
     using Cube = std::vector<PTRef>;
 
     // Proof obligation: prove `cube` is not reachable at `level` steps from Init.
@@ -73,7 +71,13 @@ private:
     bool      useUnsatCoreGeneralization_{true};
     bool      minimizeRefinementPredicates_{true};
     bool      useBinaryRefinementInterpolants_{false};
-    bool      addInitialReset_{true};
+    bool      addInitialReset_{false};
+    bool      makeSimpleProperty_{false};
+
+    // When makeSimpleProperty_ is on, the original bad formula is stashed here
+    // so the SAFE invariant can be translated back (substitute p -> origBad).
+    PTRef     simplePropOrigBad_{PTRef_Undef};
+    PTRef     simplePropVar_{PTRef_Undef};
 
     // Set before calling runIC3(); setupAbstractSystem() fills these with abstract formulas.
     PTRef                init_{PTRef_Undef};
@@ -90,7 +94,6 @@ private:
 
     // --- IC3 core methods ---
 
-    void resetIC3State();
     TransitionSystemVerificationResult runIC3(bool resuming = false);
 
     void  pushFrame()          { ++depth_; }
@@ -104,7 +107,7 @@ private:
                       std::vector<PTRef> const & vars) const;
 
     bool  initIntersects(Cube const & cube) const;
-    bool  hasBadSuccessor(Cube & outCube) const;
+    bool  hasBadState(Cube & outCube) const;
     bool  hasPredecessorUnder(PTRef frameFormula, Cube const & cube,
                                Cube & outCTI) const;
     bool  hasPredecessor(Cube const & cube, unsigned level, Cube & outCTI) const;
@@ -139,6 +142,7 @@ private:
                                                       std::size_t depth) const;
     bool addPredicate(PTRef pred);
     void applyInitialReset();
+    void applySimpleProperty();
     void initializePredicates();
 
     // --- Abstract system construction ---
@@ -154,14 +158,13 @@ private:
     PTRef concreteInvariant(PTRef abstractInv) const;
 
     // --- Persistent solvers for IC3 queries ---
-    //
-    // Both solvers are seeded once with the abstract system's constant context
-    // (init_ / trans_) in initializeAbstractSystem(). New label-definition
-    // conjuncts produced by extendAbstractSystem() are asserted incrementally,
-    // outside any push scope, so they persist across queries. Each query uses
-    // push()/pop() to scope its query-specific assertions.
     std::unique_ptr<SMTSolver> initSolver_;
     std::unique_ptr<SMTSolver> transSolver_;
+
+    // Activation literal guarding concreteTrans_ in transSolver_. Asserted true
+    // for queries that need the transition relation (predecessor, propagation),
+    // and false for queries about a single state (hasBadState).
+    PTRef transAct_{PTRef_Undef};
 
     // --- Concrete CEGAR check ---
 
--- a/src/engine/IC3IA.h
+++ b/src/engine/IC3IA.h
@@ -91,7 +91,7 @@ private:
     // --- IC3 core methods ---
 
     void resetIC3State();
-    TransitionSystemVerificationResult runIC3();
+    TransitionSystemVerificationResult runIC3(bool resuming = false);
 
     void  pushFrame()          { ++depth_; }
     PTRef getF(unsigned k) const;
@@ -143,9 +143,25 @@ private:
 
     // --- Abstract system construction ---
 
+    // Number of predicates whose label-definition equalities have already been
+    // conjoined into init_/trans_/bad_. Maintained monotonically across CEGAR
+    // iterations so that extendAbstractSystem() only asserts new predicates.
+    std::size_t numAssertedPreds_{0};
+
     PTRef abstractFormula(PTRef fla) const;
-    void  setupAbstractSystem();
+    void  initializeAbstractSystem();
+    void  extendAbstractSystem();
     PTRef concreteInvariant(PTRef abstractInv) const;
+
+    // --- Persistent solvers for IC3 queries ---
+    //
+    // Both solvers are seeded once with the abstract system's constant context
+    // (init_ / trans_) in initializeAbstractSystem(). New label-definition
+    // conjuncts produced by extendAbstractSystem() are asserted incrementally,
+    // outside any push scope, so they persist across queries. Each query uses
+    // push()/pop() to scope its query-specific assertions.
+    std::unique_ptr<SMTSolver> initSolver_;
+    std::unique_ptr<SMTSolver> transSolver_;
 
     // --- Concrete CEGAR check ---
 
--- a/src/engine/IC3IA.h
+++ b/src/engine/IC3IA.h
@@ -1,0 +1,158 @@
+/*
+ * Copyright (c) 2025
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+#ifndef GOLEM_IC3IA_H
+#define GOLEM_IC3IA_H
+
+#include "Options.h"
+#include "TermUtils.h"
+#include "TransitionSystem.h"
+#include "TransitionSystemEngine.h"
+#include "utils/SmtSolver.h"
+
+#include <memory>
+#include <queue>
+#include <chrono>
+#include <unordered_map>
+#include <vector>
+
+namespace golem {
+
+/*
+ * IC3 with Implicit Predicate Abstraction (IC3IA).
+ *
+ * CEGAR loop:
+ *   1. Extract initial predicates from init and bad (their atomic sub-formulas).
+ *   2. Build abstract Boolean TransitionSystem:
+ *        stateVars  = { l_i (Bool) ↔ p_i(x##0) }
+ *        trans      = concreteT ∧ (l_i ↔ p_i(x##0)) ∧ (l_i' ↔ p_i(x##1))
+ *        init / bad = concrete init/bad with atoms replaced by labels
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

**Comment by blishko** [comment]
Thank you! I'll try to have a look soon and provide feedback.

**Comment by blishko** [comment]
Hi @gianredondi, I am looking at the code and I have a few thoughts.

Most importantly, you should enable validation of the witnesses in the tests, it is currently disabled. I tried enabling validation for the first test, but it failed, which might indicate some issues with the implementation.

While the current version is a good start, it would require some cleaning up to conform to the style of Golem's codebase. Do you want to work on that further or would be OK for me to take over?

Regarding the implementation, I am not familiar with the details of the algorithm, but my impression from the original IC3-IA paper is that you do not include the concrete transition relation when running IC3 on the abstract system. However, in your implementation, you do include the concrete transition relation as one of the conjuncts of the abstract one, which does not feel right. Can you comment on that?

**Comment by gianredondi** [comment]
> Hi @gianredondi, I am looking at the code and I have a few thoughts.
> 
> Most importantly, you should enable validation of the witnesses in the tests, it is currently disabled. I tried enabling validation for the first test, but it failed, which might indicate some issues with the implementation.

Hi @blishko! Thanks for the feedback.
I have fixed witness validation, and modified the tests. There was an issue with the addInitialReset option.

This option enables a transformation trick (also implemented in the original IC3IA) 
that replaces the initial formula of the transition system with a boolean flag ic3ia_reset. 
The transition relation is modified to: ¬reset' ∧ (¬reset ∨ Init') ∧ (reset ∨ Trans), 
and the bad formula becomes ¬reset ∧ Bad. 
This is useful because the initial formula often contains many atoms that would be mined as predicates, 
slowing down the abstraction.

The fix eliminates ic3ia_reset from the inductive invariant, and reduce by 1 the depth of cex witnesses.

More in general I am unsure of what to do in case of multi-predicate CHC. I understand that these are transformed into a transition system via the SingleLoopTransformation, but I have not figured out how to map back the inductive invariant found over the original CHC system. In such cases validation still fails.

> While the current version is a good start, it would require some cleaning up to conform to the style of Golem's codebase. Do you want to work on that further or would be OK for me to take over?

I do not have fixed plan on this, so I am perfectly fine if you want to do any cleaning! 

> Regarding the implementation, I am not familiar with the details of the algorithm, but my impression from the original IC3-IA paper is that you do not include the concrete transition relation when running IC3 on the abstract system. However, in your implementation, you do include the concrete transition relation as one of the conjuncts of the abstract one, which does not feel right. Can you comment on that?

Yes, the original paper is using a transition formula Trans(\bar{X}, X') plus additional EQ formulas to encode the abstraction. But actually, it is equivalent to use just the H formulas (the ones connecting labels with the original predicates) and the concrete Trans(X, X'). I found easier to implement IC3IA in this way, and in my previous tests with other implementation the difference was negligible. Let me know if you want to discuss this further!

**Comment by blishko** [comment]
Hi @gianredondi, thanks for the updates!
I did a quick evaluation on [transition system benchmarks over LRA](https://github.com/chc-comp/chc-comp21-benchmarks/tree/main/LRA-TS).
Good news is that the implementation did not return any incorrect result.
Not so good news is that the engine is not very efficient compared to the other engines, like PD-KIND and Spacer, which are closely related. Spacer on transition systems is basically IC3 on the concrete system (with specific generalization techniques, though).

Browsing through the code, I saw a couple of opportunities for optimization.

I suggest the following course of actions: I would merge your version (possibly with some small modifications) so that it is recorded in the history. Then I would take over and change the style and try to optimize the algorithm.

**Comment by blishko** [comment]
> This option enables a transformation trick (also implemented in the original IC3IA)
that replaces the initial formula of the transition system with a boolean flag ic3ia_reset.
The transition relation is modified to: ¬reset' ∧ (¬reset ∨ Init') ∧ (reset ∨ Trans),
and the bad formula becomes ¬reset ∧ Bad.
This is useful because the initial formula often contains many atoms that would be mined as predicates,
slowing down the abstraction.

Is the only goal to basically shift the traces of the system one step in order to *not* mine predicates from the initial states?
If so, why just not implement directly that change (mine predicates from initial states only if some flag is set) instead of modifying the transition system?

> More in general I am unsure of what to do in case of multi-predicate CHC. I understand that these are transformed into a transition system via the SingleLoopTransformation, but I have not figured out how to map back the inductive invariant found over the original CHC system. In such cases validation still fails.

You should not need to do anything.
In this case, you only need to return a correct certificate for the transition system you get on input. The rest of the framework will take care of the rest of backtranslation.
Do you have a failing case? Can you share it?

> Yes, the original paper is using a transition formula Trans(\bar{X}, X') plus additional EQ formulas to encode the abstraction. But actually, it is equivalent to use just the H formulas (the ones connecting labels with the original predicates) and the concrete Trans(X, X'). I found easier to implement IC3IA in this way, and in my previous tests with other implementation the difference was negligible. Let me know if you want to discuss this further!

Is this described in some other paper? That reminds me, can you add links to the relevant papers in the source code?
Can you add comments to the options of ic3ia to explain how they modify the algorithm?

**Comment by gianredondi** [comment]
Hi, I also did a quick evaluation on 98 LIA CHC taken from chc-comp-25, here is the benchmark set: 
[lia_bench.txt](https://github.com/user-attachments/files/27115728/lia_bench.txt). 

With a timeout of 30 seconds, the IC3IA engine solved 71 out of 98 instances, while Spacer solved 84 out of 98. 
Overall, Spacer tends to be faster: performance is comparable on easier instances, but Spacer performs better on harder ones.

IC3IA uniquely solves 1 benchmark that Spacer does not, whereas Spacer uniquely solves 13 instances.

I am interested in understanding the reasons behind this performance gap, 
but I am unsure when I will have time to go more in the details of this.

Additionally, 14 benchmarks in this set are safe multi-predicate CHCs, and in all these cases IC3IA validation fails. I am attaching this subset as well: [validation_fails.txt](https://github.com/user-attachments/files/27116033/validation_fails.txt)

I will soon update the README with a brief description of IC3IA, a link to the paper, and an overview of the available options.

**Comment by blishko** [comment]
> Additionally, 14 benchmarks in this set are safe multi-predicate CHCs, and in all these cases IC3IA validation fails. I am attaching this subset as well: [validation_fails.txt](https://github.com/user-attachments/files/27116033/validation_fails.txt)

Actually, the benchmark `vmt-chc-benchmarks/lustre/metros_1_e2_1102_000.smt2` has only one predicate. The benchmark defines a transition system, one predicate for the invariant + (init, tr, bad) clauses. So no transformations are taking place, yet the witness is still not validated.
I will have to take a closer look to understand where the problem is.

**Comment by gianredondi** [comment]
I realized that my IC3IA implementation was not exploiting incrementality across refinement steps. 
The IC3IA paper states that since the set of predicates only grows monotonically across refinements, 
the abstract transition relation becomes strictly stronger at each step,
and this means that any blocking clause proven inductive w.r.t. the previous predicate set remains inductive after refinement, therefore the already built frames should be preserved rather than discarded.

I refactored the implementation to exploit this incrementality and also 
to use persistent solver instances for queries.

I re-run the evaluation on 198 LIA benchmarks from CHC-COMP-25 (120s timeout) 
[benchlist-198.txt](https://github.com/user-attachments/files/27306255/benchlist-198.txt):

| Metric | IC3IA | Spacer |
|---|---|---|
| Solved | 141/198 (71.2%) | 172/198 (86.9%) |
| Avg time (solved) | 0.851s | 2.930s |

Now IC3IA is **3.4× faster** on instances it solves. 
There are still 20 validation fails though: 
[validation-failures.txt](https://github.com/user-attachments/files/27306261/validation-failures.txt)

**Comment by blishko** [comment]
> I refactored the implementation to exploit this incrementality and also to use persistent solver instances for queries.

Nice! Indeed, being able to re-use the knowledge gained even after refinements of the predicates is quite important for performance!

I am running my own evaluation at the moment.

> Avg time (solved)	0.851s	2.930s

Is this average time over the commonly solved instances? Or average time over the instances solved per engine?

**Comment by blishko** [comment]
Here is a minimized example where the validation fails at the moment:
```
(set-logic HORN)
(declare-fun f (Int) Bool)
(declare-fun l (Int) Bool)
(assert (forall ((H Int)) (=> (f H) (f 0))))
(assert (forall ((G Int)) (f 0)))
(assert (forall ((B Int)) (=> (f B) (l B))))
(assert (forall ((B Int)) (=> (and (not (= 0 B)) (l B)) false)))
(check-sat)
```

I am investigating.

**Comment by gianredondi** [comment]
> Is this average time over the commonly solved instances? Or average time over the instances solved per engine?

This was average time over instances solved per engine, not over commonly solved, and in fact it was probably misleading... IC3IA is now faster then the previous version, but still slower then Spacer on the commonly solved instances.

**Comment by blishko** [comment]
Here is a test you can add. The validation currently fails:

```
TEST_F(IC3LIATest, test_IC3IA_single_feasible_state) {
    Options options;
    options.addOption(Options::LOGIC, "QF_LIA");
    options.addOption(Options::COMPUTE_WITNESS, "true");
    SymRef s1 = mkPredicateSymbol("s1", {intSort()});
    PTRef current = instantiatePredicate(s1, {x});
    PTRef next    = instantiatePredicate(s1, {xp});
    // Init: x = 0; Trans: x' = 0; Bad: x > 0
    std::vector<ChClause> clauses{
        {
            ChcHead{UninterpretedPredicate{next}},
            ChcBody{{logic->mkEq(xp, zero)}, {}}
        },
        {
            ChcHead{UninterpretedPredicate{next}},
            ChcBody{{logic->mkEq(xp, zero)}, {UninterpretedPredicate{current}}}
        },
        {
            ChcHead{UninterpretedPredicate{logic->getTerm_false()}},
            ChcBody{{logic->mkGt(x, zero)},
                    {UninterpretedPredicate{current}}}
        }
    };
    IC3IA engine(*logic, options);
    solveSystem(clauses, engine, VerificationAnswer::SAFE, true);
}
```

From what I have seen, the engine computes an invariant that is just `true`.
However, that is not a safe invariant.

**Comment by blishko** [comment]
I had another look at the problematic example.
I suspect the problem is following:
When the algorithm detects that bad state has no more predecessors and it detects a fix point in the frame, it assumes the abstract frame is the safe inductive invariant.
However, because you include the concrete transition and bad state in the check together with the abstract ones, it affects this check for predecessor.
It can happen, as in this example, that there would be a predecessor in the abstract system, but the concrete portion of the check makes the query unsatisfiable.

I am not sure what the correct fix is. You have to compute the inductive invariant differently, not just take it from the abstract system. But I don't know how exactly it needs to be strengthened.

**Comment by gianredondi** [comment]
Hi,
some improvements. In particular:

- As you spotted, in the main IC3 loop I was checking whether the successor of a state could be bad rather than the state itself. Fixing this (`hasBadSuccessor` → `hasBadState`) caused validation to be correct (surprisingly this never produced a wrong overall result and I didn't catch it earlier).
- A single persistent solver is now reused across CEGAR rounds, with activation literals selecting init/trans. This improved the incrementality of the approach.

I also did some small changes in the refinement procedures and cleaned some redundant checks in the code.

### Evaluation

I re-ran on 183 linear CHCs ([benchlist_183.txt](https://github.com/user-attachments/files/27549428/benchlist_183.txt))  from CHC-COMP25 (I realized 15 entries in the 198 list I had been using were non-linear, so I dropped them). 120s timeout per benchmark.

|         | IC3IA | Spacer |
| ------- | ----: | -----: |
| SAFE    |    93 |    100 |
| UNSAFE  |    60 |     57 |
| TIMEOUT |    30 |     26 |
| **Solved** | **153** | **157** |

- **Soundness**: 0 disagreements, validation always successful. 
- Uniquely solved: IC3IA 4 (1 SAFE, 3 UNSAFE), Spacer 8 (all SAFE).
- Time on the 149 commonly-solved benchmarks: mean **1.64s** (IC3IA) vs **1.02s** (Spacer); medians **0.030s** vs **0.030s**.



Next I will restructure the `initial-reset` option as you suggested so it doesn't modify the TS directly (just start the algo without predicates from init).

**Comment by blishko** [comment]
Looks very good!
After restructuring the `initial-reset` option, I'll take another look and we should be able to merge then!

File: src/engine/IC3IA.h [hunk 1]
```diff
@@ -0,0 +1,168 @@
+/*
+ * Copyright (c) 2025
+ *
```
blishko, line 3: @gianredondi, would you like to add your name or pseudonym here?
Or do you prefer not to?

gianredondi: It is okay like this, thanks

blishko (approved):
Thanks again!

